PR #2, opened 2019-03-27
Service Plan create_json_schema

slemrmartin:
survey_spec converted to data-driven-forms format

http://data-driven-forms.surge.sh/renderer/form-schemas

PR #4, opened 2019-03-28
Service Plan create_json_schema

slemrmartin:
survey_spec converted to data-driven-forms format

http://data-driven-forms.surge.sh/renderer/form-schemas

Replaces #2, which was commited to upstream

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -1,0 +1,20 @@
+require "bundler/setup"
+require "topological_inventory/ansible_tower/collector"
+
+RSpec.configure do |config|
+  # Enable flags like --only-failures and --next-failure
+  config.example_status_persistence_file_path = ".rspec_status"
+
+  config.expect_with :rspec do |c|
+    c.syntax = :expect
+  end
+end
+
+spec_path = File.dirname(__FILE__)
+Dir[File.join(spec_path, "support/**/*.rb")].each { |f| require f }
+
+#
+# You can add local requires to /lib/mock/require.dev.rb
+#
+require_dev_path = File.join(spec_path, "../lib", "require.dev.rb")
+require require_dev_path if File.exist?(require_dev_path)
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -12,9 +12,3 @@ end
 
 spec_path = File.dirname(__FILE__)
 Dir[File.join(spec_path, "support/**/*.rb")].each { |f| require f }
-
-#
-# You can add local requires to /lib/mock/require.dev.rb
-#
-require_dev_path = File.join(spec_path, "../lib", "require.dev.rb")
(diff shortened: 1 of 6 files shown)


slemrmartin:
@agrare only small changes against #2 there, pls merge

spec/spec_helper.rb
```diff
@@ -0,0 +1,20 @@
...
+end
+
+spec_path = File.dirname(__FILE__)
+Dir[File.join(spec_path, "support/**/*.rb")].each { |f| require f }
+
+#
+# You can add local requires to /lib/mock/require.dev.rb
+#
+require_dev_path = File.join(spec_path, "../lib", "require.dev.rb")
+require require_dev_path if File.exist?(require_dev_path)
```
agrare: @slemrmartin this is handled by our bundler plugin in the main Gemfile in a much more flexible way, please remove this

slemrmartin: @agrare According to readme, bundler-inject is able override gems, how to add local `require "something"` like "pry-byebug" by this gem?. I don't want to explicitly write "require 'pry-byebug'" every time I need to debug something

agrare: @slemrmartin bundler-inject really breaks down into two things
1. It pulls in extra gemfiles (e.g. from .bundler.d/*.rb and ~/.bundler.d/*`
2. It adds the extra DSL methods `override_gem` and `ensure_gem` which can be used in those extra gemfiles

But that doesn't mean that you have to only use `override_gem`.  You can just `gem 'pry-byebug'` in your bundler.d/ gemfile.

Personally I have `ensure_gem 'pry-byebug'` in my global `~/.bundler.d/pry.rb` file.  This works like `gem 'pry-byebug'` if pry-byebug isn't already in the applications Gemfile, or if it is it works like `override_gem 'pry-byebug'` and will override the other gem options.

slemrmartin: @agrare yes I added `ensure_gem 'pry-byebug' `, but it's only code for bundler, doesn't add `require` for runtime as this code does (at least on my pc debugging doesn't work without `require`).

agrare: That sounds like an issue we should fix in bundler-inject rather than add a completely new thing, did you try adding a `:require => 'pry-byebug'` to the ensure_gem?

cc @Fryguy

slemrmartin: @agrare yes, but it doesn't work, I think it defines dependencies between gems, it doesn't load classes at runtime.

slemrmartin: @agrare it's not needed by this PR, removing. This feature is missing in all repos

agrare: @slemrmartin can you open an issue on https://github.com/ManageIQ/bundler-inject then, it should work.

agrare (approved):

PR #5, opened 2019-03-28
Removing runing script's defaults

slemrmartin:
just removing defaults with credentials, accidentally committed (but this host requires vpn and doesn't contain sensitive data, so it's not security incident)

cc @Ladas


slemrmartin:
@Ladas pls merge

Ladas (approved):
:+1:

PR #7, opened 2019-05-22
Add metrics to the collector

slemrmartin:
This adds an in-process metrics server using the prometheus exporter
gem and records an error metric each time an exception is caught
in the collector loop.

### DIFF
--- a/lib/topological_inventory/ansible_tower/collector.rb
+++ b/lib/topological_inventory/ansible_tower/collector.rb
@@ -11,13 +11,14 @@ module TopologicalInventory::AnsibleTower
     require "topological_inventory/ansible_tower/collector/service_catalog"
     include TopologicalInventory::AnsibleTower::Collector::ServiceCatalog
 
-    def initialize(source, tower_hostname, tower_user, tower_passwd)
+    def initialize(source, tower_hostname, tower_user, tower_passwd, metrics)
       super(source, :default_limit => 5)
 
       self.connection_manager = TopologicalInventory::AnsibleTower::Connection.new
       self.tower_hostname = tower_hostname
       self.tower_user = tower_user
       self.tower_passwd = tower_passwd
+      self.metrics = metrics
     end
 
     def collect!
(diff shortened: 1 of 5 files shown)


slemrmartin:
@carbonin pls review

lib/topological_inventory/ansible_tower/collector.rb
```diff
@@ -86,8 +88,9 @@ def collector_thread(connection, entity_type)
       # connection.api.jobs.all
       # connection.api.workflow_jobs.all
     rescue => e
+      metrics.record_error
```
slemrmartin, line 91: @carbonin It should be probably there in all collectors, right?

lib/topological_inventory/ansible_tower/collector.rb
```diff
@@ -86,8 +88,9 @@ def collector_thread(connection, entity_type)
...
       logger.error("Error collecting :#{entity_type}, message => #{e.message}")
-      raise e
+      logger.error(e)
```
carbonin, line 93: Why are you making this change here?

lib/topological_inventory/ansible_tower/collector.rb
carbonin: Are any of the collectors missing this?

slemrmartin: all of them. They call record_errors only for targeted refresh, not in full_refresh's thread

lib/topological_inventory/ansible_tower/collector.rb
slemrmartin: because I want to unify behavior with other collectors

lib/topological_inventory/ansible_tower/collector.rb
carbonin: Okay, then make the change. My goal was to just make the metrics available in all the components then the people who are most familiar with the source can add whatever metrics they think would be useful.

Also, since you mentioned threads, I'm not sure the metrics object is thread safe. Maybe making it thread safe would be a candidate for a separate gem. cc @agrare

carbonin: Or maybe we can contribute that back to prometheus exporter itself

PR #8, opened 2019-06-12
Fixed collections validations, updated save&sweep

slemrmartin:
Collections validations are fixed according to ingress api validations,
save & sweep methods provided with inventory_name and schema_name


slemrmartin:
@Ladas pls review

Ladas (approved):
:+1:

PR #9, opened 2019-06-26
Run the collector in a loop

agrare:
Previously the collector would run once and exit, this changes it to
collect in a loop for a configurable frequency.


slemrmartin (approved):
:+1:

PR #10, opened 2019-06-26
Add a scaffold operations worker for ansible tower

agrare:
Adds a simple worker that waits for operations commands on the operations kafka topic.

### DIFF
--- a/lib/topological_inventory/ansible_tower/operations/worker.rb
+++ b/lib/topological_inventory/ansible_tower/operations/worker.rb
@@ -1,0 +1,57 @@
+require "topological_inventory/ansible_tower/logging"
+
+module TopologicalInventory
+  module AnsibleTower
+    module Operations
+      class Worker
+        include Logging
+
+        def initialize(messaging_client_opts = {})
+          self.messaging_client_opts = default_messaging_opts.merge(messaging_client_opts)
+        end
+
+        def run
+          # Open a connection to the messaging service
+          require "manageiq-messaging"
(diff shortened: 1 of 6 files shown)


lib/topological_inventory/ansible_tower/operations/worker.rb
```diff
@@ -0,0 +1,57 @@
...
+          end
+        ensure
+          client&.close
+        end
+
+        private
+
+        attr_accessor :messaging_client_opts
+
+        def process_message(message)
```
slemrmartin, line 31: it can call `Processor.process` in this PR, it's generic code too

slemrmartin (approved):
:+1: Nice backbone, it would be great to extract this code to common (ingress-api-client?), most of the code can be shared with openshift-operations (+will be shared with azure?)

agrare:
> it would be great to extract this code to common (ingress-api-client?)

Eh I would prefer not to put any operations code in the ingress-api-client.  I'm not a huge fan of some of the collector common code there in the first place and this has even less to do with ingress-api.  If anything collector-common and operations-common gems would be better.

lib/topological_inventory/ansible_tower/operations/worker.rb
agrare: :+1: added with specs

Fryguy:
Nice.  +1 for the shared code idea, but not in a client...we need some sort of shared topo gem...perhaps topo core?  Either way, that's for the future.

agrare:
Yeah actually topo-core could be a good place :+1:

PR #11, opened 2019-07-02
Fix the env vars for authentication info

agrare:
This had ENDPOINT_USER/ENDPOINT_PASSWORD when it should have been
AUTH_USERNAME/AUTH_PASSWORD leading to:
```
Error: option --user must be specified.
Try --help for help.
```


slemrmartin (approved):
:+1: I wanted to do the same PR right now :)

PR #12, opened 2019-07-10
Ordering jobs by Operations worker

slemrmartin:
Operation receives service_instance(==tower job) and user form inputs (if job has survey). Then it launches job and waits in thread for finish (1800 seconds max) and for collector for url (1800 seconds max)

### DIFF
--- a/lib/topological_inventory/ansible_tower/operations/core/ansible_tower_client.rb
+++ b/lib/topological_inventory/ansible_tower/operations/core/ansible_tower_client.rb
@@ -57,6 +57,7 @@ module TopologicalInventory
             # This means that api_client:job_template.launch() called job.find(nil), which returns list of jobs
             # => status error was returned, but api_client doesn't return errors
             raise ::AnsibleTowerClient::ResourceNotFoundError, "Job not found" if job.respond_to?(:count)
+
             job
           end
 
@@ -74,7 +75,8 @@ module TopologicalInventory
               end
 
               break if (count += 1) >= timeout_count
-              sleep(SLEEP_POLL) #seconds
+
+              sleep(SLEEP_POLL) # seconds
             end
             job
           end
@@ -115,8 +117,8 @@ module TopologicalInventory
           end
 
           def ansible_tower
-            @connection ||= connection_manager.connect(
-              default_endpoint.host, authentication.username, authentication.password, verify_ssl: verify_ssl_mode
+            @ansible_tower ||= connection_manager.connect(
+              default_endpoint.host, authentication.username, authentication.password, :verify_ssl => verify_ssl_mode
             )
           end
         end
--- a/lib/topological_inventory/ansible_tower/operations/core/ansible_tower_client.rb
+++ b/lib/topological_inventory/ansible_tower/operations/core/ansible_tower_client.rb
@@ -1,0 +1,126 @@
+require "topological_inventory/ansible_tower/logging"
+require "topological_inventory/ansible_tower/connection"
+require "topological_inventory/ansible_tower/operations/core/sources_api_client"
+require "topological_inventory/ansible_tower/operations/core/topology_api_client"
+
+module TopologicalInventory
+  module AnsibleTower
+    module Operations
+      module Core
+        class AnsibleTowerClient
+          include Logging
+          include Core::TopologyApiClient
+
+          attr_accessor :connection_manager
+
+          SLEEP_POLL = 10
+          POLL_TIMEOUT = 1800
+
+          def initialize(source_id, task_id, identity = nil)
+            self.identity   = identity
+            self.source_id  = source_id
+            self.task_id    = task_id
+
+            self.connection_manager = TopologicalInventory::AnsibleTower::Connection.new
+          end
+
+          # Format of order params (Input for Catalog - created by Collector, Output is produced by catalog - input of this worker)
+          #
+          # @example:
+          #
+          # * Input (ServicePlan.create_json_schema field)(created by lib/topological_inventory/ansible_tower/parser/service_plan.rb)
+          #     {"schema":
+          #       {"fields":[
+          #         {"name":"providerControlParameters", ... },
+          #         {"name":"NAMESPACE", "type":"text", ...},
+          #         {"name":"MEMORY_LIMIT","type":"text","default":"512Mi","isRequired":true,...},
+          #         {"name":"POSTGRESQL_USER",type":"text",...},
+          #         ...
+          #        ]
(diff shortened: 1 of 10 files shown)


slemrmartin:
@agrare pls review

lib/topological_inventory/ansible_tower/operations/core/ansible_tower_client.rb
```diff
@@ -0,0 +1,129 @@
...
+          #       ...
+          #     }"
+          def order_service_plan(job_template_id, order_params)
+            job_template = ansible_tower.api.job_templates.find(job_template_id)
+
+            job = job_template.launch(job_values(order_params))
+
+            # This means that api_client:job_template.launch() called job.find(nil), which returns list of jobs
+            # => status error was returned, but api_client doesn't return errors
+            raise ::AnsibleTowerClient::ResourceNotFoundError, "Job not found" if job.respond_to?(:count)
```
agrare, line 59: Wow that's nasty, something we can fix in the client gem (not holding this up, just for the future)?

slemrmartin: yes, I suppose there should be more work on client, because its support for api v2 is limited

slemrmartin: it works like `find(nil)` launches the same code as `all()`

lib/topological_inventory/ansible_tower/operations/core/sources_api_client.rb
```diff
@@ -0,0 +1,85 @@
+require "sources-api-client"
+
+module TopologicalInventory
+  module AnsibleTower
+    module Operations
+      module Core
+        class SourcesApiClient < ::SourcesApiClient::ApiClient
+          INTERNAL_API_PATH = '//internal/v1.0'.freeze
```
agrare, line 8: Why two leading `/` ?

slemrmartin: I used the same style as sources-client does: https://github.com/ManageIQ/sources-api-client-ruby/blob/master/lib/sources-api-client/configuration.rb#L133
...but it's probably not necessary

lib/topological_inventory/ansible_tower/operations/core/ansible_tower_client.rb
```diff
@@ -0,0 +1,129 @@
+require "topological_inventory/ansible_tower/logging"
+require "topological_inventory/ansible_tower/connection"
+require "topological_inventory/ansible_tower/operations/core/sources_api_client"
+require "topological_inventory/ansible_tower/operations/core/topology_api_client"
+
+module TopologicalInventory
+  module AnsibleTower
+    module Operations
+      module Core
+        class AnsibleTowerClient
```
agrare, line 10: Calling this AnsibleTowerClient is a little confusing and most of this looks like it can just live with the order code

lib/topological_inventory/ansible_tower/operations/core/ansible_tower_client.rb
```diff
@@ -0,0 +1,129 @@
...
+          include Logging
+          include Core::TopologyApiClient
```
agrare, line 12: Generally prefer a class with class methods over mixin in a module

lib/topological_inventory/ansible_tower/operations/core/ansible_tower_client.rb
```diff
@@ -0,0 +1,129 @@
...
+
+          SLEEP_POLL = 10
+          POLL_TIMEOUT = 1800
+
+          def initialize(source_id, task_id, identity = nil)
+            self.identity   = identity
+            self.source_id  = source_id
+            self.task_id    = task_id
+
+            self.connection_manager = TopologicalInventory::AnsibleTower::Connection.new
```
agrare, line 24: This isn't really a connection manager like openshift has (and would argue it doesn't need to be, ansible_tower is much less complex than what openshift has to deal with).

lib/topological_inventory/ansible_tower/operations/core/sources_api_client.rb
```diff
@@ -0,0 +1,85 @@
...
+
+          def initialize(identity = nil)
+            super(::SourcesApiClient::Configuration.default)
+            self.identity = identity
+            self.api = init_default_api
+          end
```
agrare, line 14: I get what you're doing here but subclassing a client gem here seems weird.  This is something that all our services currently do so if we're going to make it simpler lets do it either in the client gem or in a more common location

lib/topological_inventory/ansible_tower/operations/core/topology_api_client.rb
```diff
@@ -0,0 +1,28 @@
...
+
+          def update_task(task_id, state:, status:, context:)
+            task = TopologicalInventoryApiClient::Task.new("state" => state, "status" => status, "context" => context.to_json)
+            topology_api_client.update_task(task_id, task)
+          end
+
+          def svc_instance_url(service_instance)
+            rest_api_path = '/service_instances/{id}'.sub('{' + 'id' + '}', service_instance&.id.to_s)
+            topology_api_client.api_client.build_request(:GET, rest_api_path).url
+          end
```
agrare, line 23: Even this as a class not a mixin seems like overkill for these two methods

lib/topological_inventory/ansible_tower/operations/processor.rb
```diff
@@ -1,25 +1,133 @@
...
+          self.model           = model
+          self.method          = method
+          self.params          = payload["params"]
+          self.identity        = payload["request_context"]
         end
 
         def process
-          # TODO: handle the operation
+          logger.info("Processing #{model}##{method} [#{params}]...")
+          result = order_service(params)
```
agrare, line 31: We're going to need to handle more operations in the future so dynamically building the method instead of just calling order for any message received would be much better

agrare (approved):
:+1: lets get this in and iterate on it

lib/topological_inventory/ansible_tower/operations/core/sources_api_client.rb
slemrmartin: ok. That was the easiest way to do thread-safe switch between public and internal api calls

slemrmartin:
@agrare I chose some of these non-ideal constructions just for the reason to have the code as much similar to openshift operations, so refactoring to common code would be simpler

PR #14, opened 2019-07-16
Create ServicePlan for each ServiceOffering

slemrmartin:
To be able to order `job_template` (service_offering), `service_plan` must always exist.

Is it true? @gmcculloug

### DIFF
--- a/lib/topological_inventory/ansible_tower/parser/service_plan.rb
+++ b/lib/topological_inventory/ansible_tower/parser/service_plan.rb
@@ -2,7 +2,8 @@ module TopologicalInventory::AnsibleTower
   class Parser
     module ServicePlan
       def parse_service_plan(template, survey_spec_hash)
-        return if survey_spec_hash.blank?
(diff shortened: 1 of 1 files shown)


agrare:
Hey @slemrmartin we had talked about adding an API to order the service_offering and pass in an optional service_plan to handle this.  Ordering the offering makes more intuitive sense to me since the plan is just a set of parameters.

slemrmartin:
@agrare yes, it makes more sense. So does it worth to merge this PR now? I haven't see catalog, but I expect job_templates without survey are invisible or causes error in catalog now.

slemrmartin:
talked to @Hyperkid123, it'll cause error now, so at least as temporary solution it should be merged

agrare:
If service_offerings without service_plans cause errors in catalog that is a bug they should fix regardless of which API they call to order.
I'd rather put the time into adding the API for service_offering#order instead of merging this then backing it out.

PR #16, opened 2019-08-01
Tower URL to ServiceInstance (external_url)

slemrmartin:
Added `external_url` to `service_instances` when collecting jobs

### DIFF
--- a/lib/topological_inventory/ansible_tower/parser.rb
+++ b/lib/topological_inventory/ansible_tower/parser.rb
@@ -8,11 +8,23 @@ module TopologicalInventory::AnsibleTower
     include TopologicalInventory::AnsibleTower::Parser::ServicePlan
     include TopologicalInventory::AnsibleTower::Parser::ServiceOffering
 
+    def initialize(tower_host:)
+      super()
+
+      host_url = tower_host
+      host_url = "https://#{host_url}" unless host_url =~ %r{\Ahttps?:\/\/}
+      self.tower_host = host_url
+    end
+
     def parse_base_item(entity)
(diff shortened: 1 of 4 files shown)


slemrmartin:
@agrare pls :eyes:

lib/topological_inventory/ansible_tower/parser.rb
```diff
@@ -8,11 +8,23 @@ class Parser < TopologicalInventoryIngressApiClient::Collector::Parser
     include TopologicalInventory::AnsibleTower::Parser::ServicePlan
     include TopologicalInventory::AnsibleTower::Parser::ServiceOffering
 
+    def initialize(tower_host:)
+      super()
+
+      host_url = tower_host
+      host_url = "https://#{host_url}" unless host_url =~ %r{\Ahttps?:\/\/}
```
agrare, line 15: I think this would be a little cleaner with something like `uri = URI(tower_host); uri.scheme ||= "https"; host_url = uri.to_s` but this should do for now

lib/topological_inventory/ansible_tower/parser/service_instance.rb
```diff
@@ -3,14 +3,16 @@ class Parser
     module ServiceInstance
       def parse_service_instance(job_hash)
         job = job_hash[:job]
+
+        external_url = "#{self.tower_host}#{job.url}"
```
agrare, line 7: Depending on what the url looks like something like a URI.join or a File.join is probably a little safer so we don't assume a leading or trailing slash

agrare (approved):
Couple minor things but looks good to me

PR #22, opened 2019-08-14
Changed dockerfile's workdir

slemrmartin:
Workdir has to be compatible with SourceType.name to be able to mount config maps and secrets

**dependent** https://github.com/ManageIQ/topological_inventory-orchestrator/pull/31

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,7 +10,7 @@ RUN yum -y install --setopt=tsflags=nodocs \
                    && \
     yum clean all
 
-ENV WORKDIR /opt/ansible_tower-collector/
+ENV WORKDIR /opt/ansible-tower-collector/
 WORKDIR $WORKDIR
 
 COPY Gemfile $WORKDIR


slemrmartin:
@carbonin please review

bdunne:
> Workdir has to be compatible with SourceType.name to be able to mount config maps and secrets

Why? ConfigMaps and Secrets can be mounted wherever you want them.

carbonin:
@bdunne I think he's mapping the directory from the image name so that the logic can be generic in the orchestrator.

I haven't gotten a chance to look at the big orchestrator PR yet though so I was holding off on this one.

carbonin:
But honestly there's no harm in changing this if it's inconsistent with the rest of the collector images.

slemrmartin:
@bdunne as @carbonin wrote, mounting paths logic is generic, I didn't want to write any `if ansible-tower then ansible_tower`. And this is the only inconsistency among supported collectors.

bdunne:
> @bdunne as @carbonin wrote, mounting paths logic is generic, I didn't want to write any `if ansible-tower then ansible_tower`. And this is the only inconsistency among supported collectors.

:+1: Thanks, I just wanted to confirm that.

PR #25, opened 2019-08-23
Extracting operations to providers-common gem

slemrmartin:
Extracting all common classes and methods to `topological_inventory-providers-common` gem

All operation workers have the same workflow:

- load service_offering(service_plan)
- provision service
- wait for complete
- wait for collector to get API URL
- update Task

---

- [ ] **depends on**: https://github.com/ManageIQ/topological_inventory-providers-common/pull/5


slemrmartin:
topological-api and sources-api have to stay in Gemfile, because it's not possible to add them to gemspec now.

@agrare please :eyes: 
cc @Ladas

slemrmartin:
cc @abellotti - I moved significant part of operations to common gem, the same will be possible for Source.availability_check

agrare:
@slemrmartin I don't like how specific providers-common/operations is to service ordering.  I think we need to look at what we want to extract for a general operations worker and start from there because what is there currently doesn't look right.

slemrmartin:
@agrare actually it isn't, I'm rebasing your PR so ordering is one from possible operations. Btw I haven't push regarding PR for providers-common, so you cannot know how it looks :)) [EDIT - PR is ready]
You'll need sources-api/topological-api in most of operations, right?. You'll need endpoint client (in some form) in all of your operations (that can be generalized - but - we have no other operations now - and changing interface class is not as big deal in that case). You'll need shared processor and worker parts also.
We can discuss about it, but I think better on provider-common's PR on which this one depends.
I agree we should speak about operations at all, what (and if) is there planned for operations to the future

slemrmartin:
cc @Ladas

Ladas (approved):
:+1: looks good

slemrmartin:
Although we started with 2 kinds of operation workers, only ansible tower is used, so this PR is not much useful, closing

PR #27, opened 2019-08-28
Multithreaded collector

slemrmartin:
Collector was collecting entity types sequentially (it was intended only for simpler development). 
Switching to multi-threaded


slemrmartin:
@miq-bot add-label enhancement

slemrmartin:
@Ladas pls :eyes:

Ladas (approved):
:+1: 

downside of this is that memory requirement will rise. We should monitor it and maybe do a smaller threadpool. (using e.g. parallel gem)

PR #29, opened 2019-08-30
Validation of collector input args

slemrmartin:
Input arguments of source in multisource mode are not tested in `bin script`. They should be tested before creating Thread with collector

- [x] **depends on** https://github.com/ManageIQ/topological_inventory-ingress_api-client-ruby/pull/51

### DIFF
--- a/lib/topological_inventory/ansible_tower/collectors_pool.rb
+++ b/lib/topological_inventory/ansible_tower/collectors_pool.rb
@@ -18,6 +18,16 @@ module TopologicalInventory::AnsibleTower
       File.expand_path("../../../secret", File.dirname(__FILE__))
     end
 
+    def source_valid?(source, secret)
+      missing_data = [source.source,
+                      source.host,
+                      secret["username"],
+                      secret["password"]].select do |data|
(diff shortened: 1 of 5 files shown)


slemrmartin:
@Ladas pls :eyes:

Ladas (approved):
:+1:

lib/topological_inventory/ansible_tower/collectors_pool.rb
```diff
@@ -18,6 +18,16 @@ def path_to_secrets
       File.expand_path("../../../secret", File.dirname(__FILE__))
     end
 
+    def source_valid?(source, secret)
```
Ladas, line 21: this should be extended to actually validate the credentials, correct?

slemrmartin: it validates if credentials are present, they are in `secret`. Or do you mean validations that credentials are valid? I think this is similar to Source availability checker, which implements @abellotti - possibly if it'll be in db, then that source won't be event deployed (after merge of that)

PR #31, opened 2019-09-09
Fixing Tower URL parsing in ServiceInstance

slemrmartin:
Main problem is, when creating service instance's url, it uses `url = URI("10.0.0.1")` which behaves unexpectedly, the value is in `url.path` instead of `url.host` and then `url.to_s` is wrong. 

Second problem is, that Collectors pool should add scheme from configmap

### DIFF
--- a/lib/topological_inventory/ansible_tower/collectors_pool.rb
+++ b/lib/topological_inventory/ansible_tower/collectors_pool.rb
@@ -29,7 +29,10 @@ module TopologicalInventory::AnsibleTower
     end
 
     def new_collector(source, secret)
-      TopologicalInventory::AnsibleTower::Collector.new(source.source, source.host, secret["username"], secret["password"], metrics)
+      url = URI::Generic.build(:scheme => source.scheme.to_s.strip.presence || 'https',
+                               :host   => source.host.to_s.strip,
+                               :port   => source.port.to_s.strip)
+      TopologicalInventory::AnsibleTower::Collector.new(source.source, url.to_s, secret["username"], secret["password"], metrics)
     end
   end
 end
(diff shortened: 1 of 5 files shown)


slemrmartin:
@agrare please review

slemrmartin:
cc @gmcculloug

lib/topological_inventory/ansible_tower/collectors_pool.rb
```diff
@@ -29,7 +29,10 @@ def source_valid?(source, secret)
     end
 
     def new_collector(source, secret)
-      TopologicalInventory::AnsibleTower::Collector.new(source.source, source.host, secret["username"], secret["password"], metrics)
+      url = URI::Generic.build(:scheme => source.scheme.to_s.strip.presence || 'https',
+                               :host   => source.host.to_s.strip,
```
agrare, line 33: @slemrmartin what is `.to_s.strip` buying you here?  Are you trying to make sure that it isn't `nil`?  How about the strip are we getting leading or trailing whitespace from the Settings?

slemrmartin: yes, `.to_s` ensures it isn't `nil`. 

`strip` is there because values are provided by user inputs in UI, parsed, splitted and I maybe some other operations, so I want to be sure.

slemrmartin: btw I cannot check it if it's possible to provide anything through UI, because insights-proxy is not working now due to domain change

agrare: Okay, nil is fine with URI build but I get if you want to strip the string as well.

agrare (approved):

PR #32, opened 2019-09-17
No need to create a blank service_plan anymore

agrare:
Now that we support order on the service_offering we don't need to
create a blank service_plan any longer.


slemrmartin:
:+1: LGTM

PR #33, opened 2019-09-23
If we fail to find service_instance fail the task

agrare:
The task should not be state: "completed", status: "ok" if we failed to
find the service_instance in the topology API after 30 minutes.

### DIFF
--- a/lib/topological_inventory/ansible_tower/operations/core/topology_api_client.rb
+++ b/lib/topological_inventory/ansible_tower/operations/core/topology_api_client.rb
@@ -16,11 +16,6 @@ module TopologicalInventory
             task = TopologicalInventoryApiClient::Task.new("state" => state, "status" => status, "context" => context)
             topology_api_client.update_task(task_id, task)
           end
-
-          def svc_instance_url(service_instance)
-            return if service_instance.nil?
-            topology_api_client.api_client.build_request(:GET, "/service_instances/#{service_instance.id}").url
-          end
         end
       end
     end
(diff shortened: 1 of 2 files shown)


lib/topological_inventory/ansible_tower/operations/core/topology_api_client.rb
```diff
@@ -18,8 +18,8 @@ def update_task(task_id, state:, status:, context:)
           end
 
           def svc_instance_url(service_instance)
-            rest_api_path = '/service_instances/{id}'.sub('{' + 'id' + '}', service_instance&.id.to_s)
```
agrare, line 21: Before if the service_instance was nil this would make an API call to `/service_instances/` which doesn't mke a lot of sense

lib/topological_inventory/ansible_tower/operations/core/topology_api_client.rb
```diff
@@ -18,8 +18,8 @@ def update_task(task_id, state:, status:, context:)
...
-            topology_api_client.api_client.build_request(:GET, rest_api_path).url
+            return if service_instance.nil?
+            topology_api_client.api_client.build_request(:GET, "/service_instances/#{service_instance.id}").url
```
agrare: I need to check but I think this needs to be `.external_url` not `.url`

agrare: We don't even need this method, we're given a TopologicalInventoryApiClient::ServiceInstance object as an argument which already has `#external_url` as an attribute, deleting it.

PR #35, opened 2019-09-26
Split ansible_tower_operations into ServicePlan and ServiceOffering classes

agrare:
Split the possible ansible-tower operations into different classes based
on the model the operation is being called on instead of hard-coding
"order_service".


gtanzillo (approved):

slemrmartin:
it was planned as next step after #25 in common gem

PR #37, opened 2019-10-02
Changed parsing of kafka message's event name

slemrmartin:
It seems that `headers['message_type']` has been removed


slemrmartin:
cc @Ladas

agrare (approved):
Nice find

PR #38, opened 2019-10-03
Adding support for Source.availability_checker

abellotti:
- Adding the Source.availability_check logic for AnsibleTower
- Adding spec test for the Source.availability_check


abellotti:
[WIP] for now, need to test it end to end.

agrare (approved):

PR #41, opened 2019-10-23
AppliedInventories operation

slemrmartin:
- [x] **depends on** https://github.com/ManageIQ/topological_inventory-api-client-ruby/pull/22

**dependent** https://github.com/ManageIQ/topological_inventory-api/pull/257


agrare:
@slemrmartin the name ApprovalInventories makes some assumptions about how this will be used (e.g. by catalog for the purpose of approvals).

Would something like ServiceOfferingInventories make sense?

slemrmartin:
@agrare yes and no :) there shoud be `ServiceOffering.service_offering_inventories` method, and `/service_offering/:id/service_offering_inventories` URL in API - but it invokes that it's simple `has_many` list.
I decided for `approval_inventories`, because this operation is really Approval Service specific, I doubt it'll make sense for any other service.
It won't be used by catalog, it should provide different list for user input.

Maybe I can switch back to `effective_inventories`, but it does mean nothing.

slemrmartin:
@agrare we can name it "applied_inventories". It describes the operation and is not connected to specific svc, what do you think?

agrare:
@slemrmartin either applied or effective inventories sounds good to me :+1:

lib/topological_inventory/ansible_tower/operations/service_offering.rb
```diff
@@ -2,7 +2,8 @@
 require "topological_inventory/ansible_tower/operations/core/ansible_tower_client"
 require "topological_inventory/ansible_tower/operations/core/service_order_mixin"
 require "topological_inventory/ansible_tower/operations/core/topology_api_client"
-
+require "topological_inventory/ansible_tower/operations/applied_inventories/parser"
+require "pry-byebug"
```
agrare: Don't forget to remove this

agrare (approved):
LGTM, take it out of wip when you're ready for this to get merged

lib/topological_inventory/ansible_tower/operations/service_offering.rb
slemrmartin: I'm testing and writing specs now, but..yes, thanks, it happens to me :)

slemrmartin:
@agrare it's ready now

PR #42, opened 2019-11-07
Collecting Artifacts and ExtraVars

slemrmartin:
TPINVTRY-730

Data set in playbooks byt the `set_stats` command are saved in Job as "artifacts".
On target Jobs, they are set as `extra_vars`.

`artifacts` are not propagated from Jobs to Workflows, so consumer Job connected by edge from Workflow containing Job with artifacts will fail => `set_stats` for Workflows is not applicable unlike task requested.

`extra_vars` for workflows connected by edge from job with artifacts and for their children are computed by tower and accessible by API.

### DIFF
--- a/lib/topological_inventory/ansible_tower/parser/service_instance.rb
+++ b/lib/topological_inventory/ansible_tower/parser/service_instance.rb
@@ -18,6 +18,13 @@ module TopologicalInventory::AnsibleTower
           :finished => job.finished,
           :status   => job.status
         }
+        # launch variables set either manually, by survey values or artifacts from previous job in workflow
+        extra_vars = job.extra_vars_hash
+        extra[:extra_vars] = extra_vars if extra_vars.present?
+        if job.type == 'job'
+          artifacts = job.artifacts.to_s
+          extra[:artifacts] = YAML.safe_load(artifacts) if artifacts.present? && artifacts != '{}' # Artifacts are set by set_stats:data in playbook (Jobs only)
+        end
 
         collections.service_instances.build(
           parse_base_item(job).merge(


lib/topological_inventory/ansible_tower/parser/service_instance.rb
```diff
@@ -13,6 +13,15 @@ def parse_service_instance(job_hash)
         path         = job.type == 'workflow_job' ? 'workflows' : 'jobs/playbook'
         external_url = File.join(self.tower_url, "/#/#{path}", job.id.to_s)
 
+        extra = {}
+        # launch variables set either manually, by survey values or artifacts from previous job in workflow
+        extra_vars = job.extra_vars_hash
+        extra[:extra_vars] = extra_vars if extra_vars.present?
+        if job.type == 'job'
+          artifacts = job.artifacts.to_s
+          extra[:artifacts] = YAML.safe_load(artifacts) if artifacts.present? && artifacts != '{}' # Artifacts are set by set_stats:data in playbook (Jobs only)
```
agrare, line 26: Do we need to check for `'{}'` or could we just assign an empty hash? If we need to prevent an empty hash being set I'd rather YAML load first then check for `Object#blank?`

Could this work?
```ruby
artifacts = YAML.safe_load(job.artifacts.to_s)
extra[:artifacts] = artifacts unless artifacts.blank?
```

This works for both "empty" cases e.g. an empty hash and nil
```ruby
>> artifacts = YAML.safe_load(nil.to_s)
=> nil
>> artifacts.blank?
=> true
>> artifacts = YAML.safe_load({}.to_s)
=> {}
>> artifacts.blank?
=> true
```

slemrmartin: I wanted to avoid unecessary YAML parsing because it won't be used in most of cases I guess. I expect that `YAML.safe_load` is time-consuming operation (even if it takes small amount of time for empty hash)

agrare: Do you know if they are mostly `nil` or are they mostly `"{}"`?  My only hesitation is that `{}` isn't the only "empty state" for YAML, also e.g. `--- {}\n` is valid.

slemrmartin: it's mostly `"{}"`and never (what I saw) nil. It seems like dynamically generated class `AnsibleTowerClient::Job::Artifacts`, I didn't find it in ansible tower client.

agrare: Okay, just rebase then I'm :+1:

slemrmartin: maybe @bdunne can tell us more about it

agrare (approved):

PR #43, opened 2019-11-07
Status, Started, Finished to ServiceInstance

slemrmartin:
TPINVTRY-726

Status, Started and Finished are added to `extra` (source-type specific data)

### DIFF
--- a/lib/topological_inventory/ansible_tower/parser/service_instance.rb
+++ b/lib/topological_inventory/ansible_tower/parser/service_instance.rb
@@ -13,6 +13,12 @@ module TopologicalInventory::AnsibleTower
         path         = job.type == 'workflow_job' ? 'workflows' : 'jobs/playbook'
         external_url = File.join(self.tower_url, "/#/#{path}", job.id.to_s)
 
+        extra = {
+          :started  => job.started,
+          :finished => job.finished,
+          :status   => job.status
+        }
+
         collections.service_instances.build(
           parse_base_item(job).merge(
             :source_ref       => job.id.to_s,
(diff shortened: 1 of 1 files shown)


agrare:
@slemrmartin could we set these on the task as part of the operation instead of the service_instance?  It makes more sense for a task to have e.g. started/finished than a service_instance for other platforms (e.g. an OrchestrationStack wouldn't have a "finished" time) and I don't want catalog to depend on checking for something in `#extra` if they don't have to.

slemrmartin:
@agrare I can, but @gmcculloug wanted it in `service_instance`

agrare:
Oh?  @gmcculloug I thought we discussed that checking the task made more sense

slemrmartin:
It's defined in Jira task. So if you have discussed it, I'll change it

gmcculloug:
@agrare Correct, we did and I was ok with that.  Looks like the Jira issue was not updated.

My question is, if the service_instance is representing the Job in Tower does it belong there?  (Maybe in addition to the task?)

agrare:
I'm okay with in addition to, but I don't want you to rely on it in service_instance if we wouldn't be able to set that for any other source-type

gmcculloug:
@agrare I was thinking about how you would view these in the future and seems like data that would be included, but I was only thinking about service_instances representing Tower Jobs.  I can see how that would not work for other providers.

My suggestion would be to do whatever you feel makes the most sense for now.  Thinking that is task and we can revisit service_instance later.

agrare:
Oh this is just for display?

gmcculloug:
Yes, for Catalog order UI.

agrare:
Okay, my bad thought you were using this for waiting for it to finish

agrare (approved):

PR #45, opened 2019-11-27
[WIP] Proxy requests

slemrmartin:
"Proxy" requests are needed to on-premise tower communication through receptor.

Receptor controller doesn't support pairing requests with responses, so multithreading has to be solved with mutex on proxy side, but it'll cause slow responses.

Requests from collector can't be built as proxy requests, because Puma server on receptor-proxy app crashes during parsing (should be solved i.e. by using Go lang), so target host is provided by http header and parsed by proxy rack server.

---

https://github.com/slemrmartin/receptor-proxy/blob/master/lib/receptor_proxy/server.rb

- [ ] **depends on** https://github.com/ansible/ansible_tower_client_ruby/pull/134

### DIFF
--- a/bin/ansible_tower-collector
+++ b/bin/ansible_tower-collector
@@ -46,6 +46,12 @@ def parse_args
         :type => :string, :default => ENV["INGRESS_API"] || "http://localhost:9292"
     opt :metrics_port, "Port to expose the metrics endpoint on, 0 to disable metrics",
         :type => :integer, :default => (ENV["METRICS_PORT"] || 9394).to_i
+    opt :receptor_node_id, "ID of on-premise Receptor Node",
+        :type => :string, :default => ENV["RECEPTOR_NODE_ID"]
+    opt :proxy_scheme, "Proxy scheme",
+        :type => :string, :default => ENV["PROXY_SCHEME"] || 'http'
+    opt :proxy_host, "Proxy host",
+        :type => :string, :default => ENV["PROXY_HOST"]
   end
 end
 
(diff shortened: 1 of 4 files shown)


slemrmartin:
@gtanzillo @agrare This is a working prototype (with receptor-proxy rack server), but I'm not sure if it's worth to merge it.

Fryguy:
> because Puma server on receptor-proxy app crashes during parsing (should be solved i.e. by using Go lang)

Do we know why it crashes?  How does a go-lang rewrite solve it as opposed to just fixing whatever is the bug?

slemrmartin:
@Fryguy problem is with `URI.parse` - https://github.com/puma/puma/blob/master/lib/puma/server.rb#L556

request is parsed into env and `env['REQUEST_URI']` loses schema (https). Then 
`URI.parse('18.188.178.206:443') raises **URI::InvalidURIError: bad URI**

So it is URI bug. URI.parse('https://18.188.178.206:443') works, URI.parse('www.example.com:443') works, so it cannot parse IP address obviously. 

Anyway I chose different solution with `x-rh-endpoint` header because I wasn't able to modify proxy request targeting HTTPS (Not sure if it's even possible)

agrare:
> request is parsed into env and env['REQUEST_URI'] loses schema (https)

This sounds like the problem, do you know where this is done?

slemrmartin:
No, I didn't dig as deep yet

slemrmartin:
Will be implemented as #46, closing

PR #46, opened 2019-12-10
[WIP] Receptor API client

slemrmartin:
Receptor client for communication with Receptor API/Kafka response.

---

* Receptor API: https://github.com/RedHatInsights/platform-receptor-controller

### DIFF
--- a/lib/topological_inventory/ansible_tower/receptor/response_worker.rb
+++ b/lib/topological_inventory/ansible_tower/receptor/response_worker.rb
@@ -1,0 +1,74 @@
+module TopologicalInventory::AnsibleTower
+  module Receptor
+    class ResponseWorker
+      def self.instance
+        @instance ||= self.new
+      end
+
+      def initialize
+        @mutex   = Mutex.new
+        @started = false
+        @registered_messages = {}
+      end
+
+      def start
+        @mutex.synchronize do
+          return if @started
+          @started = true
+        end
+        Thread.new { listen }
+      end
+
+      def register_msg_id(msg_id, api_object, response_method = :response_received)
+        @registered_messages[msg_id] = { :api_object => api_object, :method => response_method }
+      end
+
+      private
+
+      def listen
+        # Open a connection to the messaging service
+        client = ManageIQ::Messaging::Client.open(default_messaging_opts)
+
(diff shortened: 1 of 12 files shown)


lib/topological_inventory/ansible_tower/receptor/response_worker.rb
```diff
@@ -0,0 +1,74 @@
+module TopologicalInventory::AnsibleTower
+  module Receptor
+    class ResponseWorker
+      def self.instance
+        @instance ||= self.new
```
agrare: Eh not a huge fan of the global singleton, could you just keep an instance in the connection and pass it wherever else is needed?

slemrmartin: ok, will do

agrare: :+1: thanks

slemrmartin: @agrare I refactored this singleton, but there is no place for instance variable. 
In this case singleton has reasonable usage, any other option doesn't improve stability or readability. So what exactly is the problem you have with singleton here?

lib/topological_inventory/ansible_tower/receptor/response_worker.rb
```diff
@@ -0,0 +1,73 @@
...
+
+      def start
+        return if started
+
+        self.started = true
+        Thread.new { listen }
+      end
+
+      def register_msg_id(msg_id, api_object, response_method = :response_received)
+        @registered_messages[msg_id] = { :api_object => api_object, :method => response_method }
```
agrare: This isn't guaranteed to be thread safe if you're using this to register from one thread while reading from it [here](https://github.com/RedHatInsights/topological_inventory-ansible_tower/pull/46/files#diff-2fde49ec8f16e8d8a0038a1e85c8dae3R44)

You could use `Concurrent::Map`

slemrmartin: good point

slemrmartin:
Replaced byt #71, closing

PR #48, opened 2019-12-17
Task's update fix, empty context not permitted

slemrmartin:
Topological API's `body_params['context']` can't be empty hash for `permit` operation

### DIFF
--- a/lib/topological_inventory/ansible_tower/operations/core/service_order_mixin.rb
+++ b/lib/topological_inventory/ansible_tower/operations/core/service_order_mixin.rb
@@ -11,7 +11,7 @@ module TopologicalInventory
             task_id, service_offering_id, service_plan_id, order_params = params.values_at(
               "task_id", "service_offering_id", "service_plan_id", "order_params")
 
-            update_task(task_id, :state => "running", :status => "ok", :context => {})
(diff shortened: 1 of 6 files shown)


slemrmartin:
Fix of #47

lib/topological_inventory/ansible_tower/operations/core/service_order_mixin.rb
```diff
@@ -11,7 +11,7 @@ def order
             task_id, service_offering_id, service_plan_id, order_params = params.values_at(
               "task_id", "service_offering_id", "service_plan_id", "order_params")
 
-            update_task(task_id, :state => "running", :status => "ok", :context => {})
+            update_task(task_id, :state => "running", :status => "ok", :context => {:info => 'Task started'})
```
agrare: @slemrmartin is context.info something that will actually be used or is this just to fix the error?
Could we just not pass any `:context` if it isn't going to contain a payload?

agrare:
Oh I was thinking just don't send any context key at all, not sure if context => nil will work

slemrmartin:
ok, it'll need different change, just a moment, :context is mandatory now

agrare:
https://github.com/RedHatInsights/topological_inventory-ansible_tower/pull/50 should get this green

PR #49, opened 2019-12-17
Improve logging for ansible tower operations

agrare:


spec/operations/worker_spec.rb
```diff
@@ -15,6 +15,8 @@
       operations_topic = "platform.topological-inventory.operations-ansible-tower"
 
       message = double("ManageIQ::Messaging::ReceivedMessage")
```
agrare, line 17: Not really a fan of this because then we have to later mock _everything_ and I don't like that adding logging breaks the tests but don't want to introduce too much change in this PR

agrare:
@gtanzillo PTAL

PR #50, opened 2019-12-17
Bump manageiq-loggers to 0.4.0

agrare:


slemrmartin (approved):
:+1:

slemrmartin:
@agrare can we remove these gem duplicates instead of maintaining the same version in provider-common and all collector repos? Or is it risky?

agrare:
@slemrmartin if managieq-loggers is used in the gem and here it should be in both, what we should do is relax the gem version requirements on providers-common so they don't have to be merged in lockstep with each other

PR #53, opened 2020-01-14
Use v2 ansible tower api in connection.rb

lpichler:


slemrmartin (approved):
c.r.c has to support Tower version >= 3.5
API v2 was introduced in Tower version 3.2.0.

:+1:

PR #55, opened 2020-01-17
[WIP] Using thread pool for memory optimization

slemrmartin:
AnsibleTower Collector uses 1 thread per entity type. In multi-source mode it uses 1 thread per source.

Which means `<number of sources> * <number of entity types>` threads.

By using ThreadPool we can significantly increase number of Sources processed by 1 collector with constant memory usage.

---

* [ ] **depends on** https://github.com/RedHatInsights/topological_inventory-providers-common/pull/11

### DIFF
--- a/lib/topological_inventory/ansible_tower/collector.rb
+++ b/lib/topological_inventory/ansible_tower/collector.rb
@@ -11,7 +11,7 @@ module TopologicalInventory::AnsibleTower
     require "topological_inventory/ansible_tower/collector/service_catalog"
     include TopologicalInventory::AnsibleTower::Collector::ServiceCatalog
 
-    def initialize(source, tower_hostname, tower_user, tower_passwd, metrics, poll_time = 60)
+    def initialize(source, tower_hostname, tower_user, tower_passwd, metrics, poll_time: 60, collector_threads: nil)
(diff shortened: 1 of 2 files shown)


slemrmartin:
@agrare this is a concept. What do you think about that?

agrare:
@slemrmartin yeah I'm not a fan of the one-thread-per-entity-type, we had to do that because of openshift and I think that got copied over when it didn't need to.

Even a thread-per-source would work for me and it might be much simpler

slemrmartin:
> @slemrmartin yeah I'm not a fan of the one-thread-per-entity-type, we had to do that because of openshift and I think that got copied over when it didn't need to.

I think I figured it out for openshift watches too. 

> 
> Even a thread-per-source would work for me and it might be much simpler

Still openshift watches have to be paused with thread pool, so it would be simpler but not as much. I'll focus on that

agrare:
> > Even a thread-per-source would work for me and it might be much simpler
> 
> Still openshift watches have to be paused with thread pool, so it would be simpler but not as much. I'll focus on that

I was thinking that openshift is different enough from the rest that we shouldn't add complexity for it to the base implementation.

slemrmartin:
closing, will be replaced by simpler PR

PR #62, opened 2020-02-12
Reduce the timout to open a connection to 10 sec

agrare:
If an attempt to connect to an Ansible Tower source hangs the default
open_timeout of 60 seconds was starving the operations worker causing
availability checks to appear to never complete.

Depends on: https://github.com/ansible/ansible_tower_client_ruby/pull/140 being released

Relates to JIRA: TPINVTRY-851


bdunne:
Can you also update the gem dependency to `~>0.21.0`?

PR #66, opened 2020-02-19
Collect join table for Service[Offering|Instance]Nodes and Credentials

lpichler:
last point (7) from https://github.com/RedHatInsights/topological_inventory-core/issues/168

@miq-bot assign @slemrmartin

### DIFF
--- a/lib/topological_inventory/ansible_tower/parser/service_instance_node.rb
+++ b/lib/topological_inventory/ansible_tower/parser/service_instance_node.rb
@@ -1,7 +1,9 @@
 module TopologicalInventory::AnsibleTower
   class Parser
     module ServiceInstanceNode
-      def parse_service_instance_node(node)
+      def parse_service_instance_node(instance_node)
+        node = instance_node[:node]
+
         service_instance_ref = node.summary_fields.job&.id&.to_s
         service_instance = lazy_find(:service_instances, :source_ref => service_instance_ref) if service_instance_ref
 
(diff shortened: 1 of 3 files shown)


lib/topological_inventory/ansible_tower/parser/service_instance_node.rb
```diff
@@ -29,6 +31,16 @@ def parse_service_instance_node(node)
             }
           )
         )
+
+        credentials = instance_node[:credentials].to_a
+        if credentials.present?
```
slemrmartin: if `credentials` are always array, then `present?` is not necessary, because `each` can handle empty array

lib/topological_inventory/ansible_tower/parser/service_offering_node.rb
```diff
@@ -26,6 +27,16 @@ def parse_service_offering_node(node)
             }
           )
         )
+
+        credentials = offering_node[:credentials].to_a
+        if credentials.present?
```
slemrmartin: same like ^^^, `present?` is not necessary

slemrmartin (changes requested):
:+1: LGTM, just simplify mentioned methods

lib/topological_inventory/ansible_tower/parser/service_instance_node.rb
lpichler: done 👍

lib/topological_inventory/ansible_tower/parser/service_offering_node.rb
lpichler: done 👍

slemrmartin (approved):

PR #71, opened 2020-03-02
On premise collector

slemrmartin:
**Issue**: https://github.com/RedHatInsights/receptor_controller-client-ruby/issues/2

Implementation of on-premise collector through Receptor network

---
- [x] **based on** https://github.com/RedHatInsights/topological_inventory-ansible_tower/pull/102

**Depends on:**

- [x]  https://github.com/RedHatInsights/receptor_controller-client-ruby/pull/5
- [x] https://github.com/RedHatInsights/receptor_controller-client-ruby/pull/12
- [x] https://github.com/RedHatInsights/topological_inventory-ansible_tower/pull/105


lib/topological_inventory/ansible_tower/collector.rb
```diff
@@ -55,7 +55,7 @@ def service_catalog_entity_types
 
     # Connection to endpoint (for each entity type the same)
     def connection_for_entity_type(_entity_type)
-      connection_manager.connect(tower_hostname, tower_user, tower_passwd)
+      connection_manager.connect(:base_url => tower_hostname, :username => tower_user, :password => tower_passwd)
```
lindgrenj6: So I see we have a receptor collector now - maybe we should rename this to cloud collector or something like that?

lib/topological_inventory/ansible_tower/receptor/async_receiver.rb
```diff
@@ -0,0 +1,59 @@
...
+        self.refresh_state_uuid = refresh_state_uuid
+        self.refresh_state_started_at = Time.now.utc
+        self.sweep_scope = Set.new
+        self.total_parts = Concurrent::AtomicFixnum.new
+        self.transformation = nil
+      end
+
+      def on_success(msg_id, entity)
+        # TODO: without tower hostname, it's not possible to construct job URL
+        parser = TopologicalInventory::AnsibleTower::Parser.new(tower_url: 'https://tower.example.com')
```
lindgrenj6: Oh this is a fun problem. We probably won't even have the hostname on receptor-enabled towers, I wonder how we'll build the URL here.

lindgrenj6 (commented):
Excited to have #102 merged then get this rebased.

lib/topological_inventory/ansible_tower/receptor/async_receiver.rb
gmcculloug: This issue was brought to my attention and something we'll need to discuss further, but should not be a blocker for this PR.  We discussed just returning the Tower job info (but not the entire URL) as well as Catalog ultimately using data returned from the Playbook set_stats to provide information to be displayed in the order.

The full URL was always more of a temporary work-around until we can get features like that into Catalog.

lib/topological_inventory/ansible_tower/collector.rb
slemrmartin: that's a good idea!

slemrmartin: @lindgrenj6 so I renamed collector to `Cloud::Collector` and Connection to `Cloud::Connection`. The PR shows them as new classes but there are no changes except the `connection_manager.connect()`

lib/topological_inventory/ansible_tower/cloud/collector.rb
```diff
@@ -0,0 +1,70 @@
...
+        self.tower_passwd   = tower_passwd
+      end
+
+      private
+
+      attr_accessor :tower_user, :tower_passwd
+
+      # Connection to endpoint (for each entity type the same)
+      def connection_for_entity_type(_entity_type)
+        connection_manager.connect(:base_url => tower_hostname, :username => tower_user, :password => tower_passwd)
```
syncrou, line 21: I'm guessing at some point this would be deprecated to use only tokens?  Or since this is the cloud connector we don't upgrade it, and there by push more users toward the more secure On Prem receptor solution.

lib/topological_inventory/ansible_tower/cloud/connection.rb
```diff
@@ -0,0 +1,34 @@
...
+        base_url = "https://#{base_url}" unless base_url =~ %r{\Ahttps?:\/\/} # HACK: URI can't properly parse a URL with no scheme
+        uri      = URI(base_url)
+        uri.path = default_api_path if uri.path.blank?
+        uri.to_s
+      end
+
+      private
+
+      def default_api_path
+        "/api/v2".freeze
```
syncrou, line 30: Interesting. Eventually this will be deprecated, but we'll have to have the `AnsibleTowerClient` modified to get that fixed.

lib/topological_inventory/ansible_tower/receptor/async_receiver.rb
```diff
@@ -0,0 +1,59 @@
...
+
+      def on_eof(_msg_id)
+        collector.async_collecting_finished(entity_type, refresh_state_uuid, total_parts.value)
+
+        collector.async_sweep_inventory(refresh_state_uuid, sweep_scope.to_a, total_parts.value, refresh_state_started_at)
+      end
+
+      private
+
+      attr_writer :connection, :entity_type, :refresh_state_uuid, :refresh_state_started_at, :sweep_scope, :total_parts
```
syncrou, line 67: Why do these have to be re declared as private `attr_writer`? Internally the `attr_reader` property is writable.

lib/topological_inventory/ansible_tower/cloud/collector.rb
slemrmartin: I'm not sure what exactly do you mean by this. Anyway this is original collector, the class is just renamed. And yes, it's for cloud connector to keep compatibility now. We can remove it later. But at this moment, most of the code is shared, so we can support both without special effort

lib/topological_inventory/ansible_tower/receptor/async_receiver.rb
slemrmartin: I didn't realize that this is redundant, :+1:

lib/topological_inventory/ansible_tower/cloud/connection.rb
slemrmartin: what's wrong with the `AnsibleTowerClient`?

lib/topological_inventory/ansible_tower/receptor/async_receiver.rb
slemrmartin: so I tried it, it's not redundant, `attr_reader` is not writable in `initialize`

syncrou: I guess I should have used the word "settable" - but potentially after being set internally those values must need to change?

Not a blocker, thanks for checking though.

syncrou (approved):
LGTM 👍 

You have any last thoughts @lindgrenj6 ?

lindgrenj6 (approved):
LGTM, can't wait to test this.

PR #75, opened 2020-03-12
Improve log messages and add request-id for tracking operations

lpichler:
TPINVTRY-838

- [x] https://github.com/ManageIQ/manageiq-loggers/pull/17

cc @slemrmartin

### DIFF
--- a/lib/topological_inventory/ansible_tower/logging.rb
+++ b/lib/topological_inventory/ansible_tower/logging.rb
@@ -15,6 +15,15 @@ module TopologicalInventory
       def logger
         TopologicalInventory::AnsibleTower.logger
       end
+
+      def log_with(request_id)
+        old_request_id = Thread.current[:request_id]
+        Thread.current[:request_id] = request_id
+
+        yield
+      ensure
+        Thread.current[:request_id] = old_request_id
+      end
     end
   end
 end
--- a/lib/topological_inventory/ansible_tower/operations/worker.rb
+++ b/lib/topological_inventory/ansible_tower/operations/worker.rb
@@ -19,11 +19,14 @@ module TopologicalInventory
 
           logger.info("Topological Inventory AnsibleTower Operations worker started...")
           client.subscribe_topic(queue_opts) do |message|
-            logger.info("Received message #{message.message}, #{message.payload}")
-            process_message(message)
+            log_with(message.payload&.fetch_path('request_context','x-rh-insights-request-id')) do
+              model, method = message.message.to_s.split(".")
+              logger.info("Received message #{model}##{method}, #{message.payload}")
+              process_message(message)
+            end
           end
         rescue => err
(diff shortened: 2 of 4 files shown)


lpichler:
@bdunne Can you review please ?

lib/topological_inventory/ansible_tower/logging.rb
```diff
@@ -15,6 +15,15 @@ module Logging
...
         TopologicalInventory::AnsibleTower.logger
       end
+
+      def log_with(request_id)
+        old_request_id = Thread.current[:request_id]
+        Thread.current[:request_id] = request_id
+
+        yield
+
+        Thread.current[:request_id] = old_request_id
```
gtanzillo, line 25: Should this be in an `ensure` block to guarantee that the original request_id is restored?

lpichler: 👍 yes, ensure block added

lpichler:
@slemrmartin also added searchable string "ServiceOffering#order" and `ServiceOffering#id` and `source_ref`

slemrmartin:
@gmcculloug this is what I was talking about last week. 
Do you have `source_ref` (of service offering and service instance) in your db, right? 
So we can connect messages this way in Kibana 

And probably we don't need request_id much..but rather more than less info

slemrmartin:
Let's wait for API's PR on CI, I'm not sure which format is the best for Kibana's filtering by message. It's more strict than expected

gmcculloug:
@mkanoor Please review

lib/topological_inventory/ansible_tower/operations/worker.rb
```diff
@@ -35,7 +38,9 @@ def run
         def process_message(message)
           Processor.process!(message)
         rescue StandardError => err
-          logger.error("#{err}\n#{err.backtrace.join("\n")}")
+          model, method = message.message.to_s.split(".")
+          task_id = message.payload&.fetch_path('params','task_id')
+          logger.error("#{model}##{method}: Task(id: #{task_id}) #{err.cause}\n#{err}\n#{err.backtrace.join("\n")}")
```
mkanoor, line 43: @lpichler can we check if this will  be handled by the insights-api-common here
https://github.com/RedHatInsights/insights-api-common-rails/blob/f0b86e82f870bb36b798a1dbccbb73edf52f8b5f/lib/insights/api/common/application_controller_mixins/exception_handling.rb#L10
If it does then we will have 2 log lines doing the same thing
Otherwise looks good

slemrmartin: @mkanoor this is not Rails app, so handler ending with `render :json` can't be used. All collector/operations repos are not using insights-api-common, but providers-common gem, where it's worth to move these common methods

slemrmartin (approved):
:+1:

PR #79, opened 2020-03-20
Operations: Waiting for failed jobs

slemrmartin:
When launching of ServiceOffering fails (in AnsibleTower), operations worker now waits for topological URL as if it succeeded.

### DIFF
--- a/lib/topological_inventory/ansible_tower/operations/core/service_order_mixin.rb
+++ b/lib/topological_inventory/ansible_tower/operations/core/service_order_mixin.rb
@@ -57,17 +57,15 @@ module TopologicalInventory
             context[:remote_status] = job.status
             task_status = client.job_status_to_task_status(job.status)
 
-            if job.status == "successful"
-              svc_instance = wait_for_service_instance(source_id, job.id)
-              if svc_instance.present?
-                context[:service_instance][:id] = svc_instance.id
-                context[:service_instance][:url] = svc_instance.external_url
-              else
-                # If we failed to find the service_instance in the topological-inventory-api
-                # within 30 minutes then something went wrong.
-                task_status = "error"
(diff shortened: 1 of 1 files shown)


lib/topological_inventory/ansible_tower/operations/core/service_order_mixin.rb
```diff
@@ -57,17 +57,15 @@ def poll_order_complete(task_id, source_id, job)
             context[:remote_status] = job.status
             task_status = client.job_status_to_task_status(job.status)
 
-            if job.status == "successful"
```
gtanzillo, line 60: This change looks straight forward. But I'm wondering if there could be an edge case where the job fails early enough that a service_instance never gets created. Do you think that could happen? If it can would it be handled, perhaps the wait would time out?

slemrmartin: in that case it raises exception: https://github.com/RedHatInsights/topological_inventory-ansible_tower/blob/master/lib/topological_inventory/ansible_tower/operations/core/ansible_tower_client.rb#L63

and this code is not processed

Ladas: before the `state: completed` was set only when finished. Now it will be set when job is running/pending which can be for hours. Didn't catalog rely on this before?

A good way to recognize the job has finished is that a finished datetime is set. (terminal states should be `successful`, `failed`, `error`, `canceled`). Job can still be hanging for a long time in a non terminal state (e.g. waiting for approval, or a job that runs for hours or days, or tower being overloaded by work)

slemrmartin: I'm waiting for `finished.present?` first, this change is after that. 
As we discussed with @Ladas, we can move waiting from operations to collector -> persister. But that's beyond scope of this PR

PR #82, opened 2020-04-07
Add external_url logging to collector

lindgrenj6:
https://projects.engineering.redhat.com/browse/TPINVTRY-881
This is the tower collector implementation for this ticket. 

Looking for input on this. currently it logs like so: 
image

Potential things I may be missing:
- should we add parameters to the logging?
- ~~maybe I could move this to the base collector class since it's just one method and it could be used across all of the collectors, that way we can just call this moethod in each `get_#{resource_type}` method call.~~ https://github.com/RedHatInsights/topological_inventory-providers-common/pull/18

cc @gmcculloug @slemrmartin 

----

EDIT:
**Depends on**: https://github.com/RedHatInsights/topological_inventory-providers-common/pull/18

### DIFF
--- a/lib/topological_inventory/ansible_tower/collector/service_catalog.rb
+++ b/lib/topological_inventory/ansible_tower/collector/service_catalog.rb
@@ -3,6 +3,7 @@ module TopologicalInventory::AnsibleTower
     module ServiceCatalog
       def get_service_credentials(connection)
         fnc = lambda do |&block|
+          log_external_url("#{connection_manager.api_url(tower_hostname)}/#{connection.api.credentials.klass.endpoint}")
           enumerator = connection.api.credentials.all(:page_size => limits[:service_credentials])
           enumerator.each do |service_credential|
             block.call(service_credential)
@@ -13,6 +14,7 @@ module TopologicalInventory::AnsibleTower
 
       def get_service_credential_types(connection)
         fnc = lambda do |&block|
+          log_external_url("#{connection_manager.api_url(tower_hostname)}/#{connection.api.credential_types.klass.endpoint}")
           enumerator = connection.api.credential_types.all(:page_size => limits[:service_credential_types])
           enumerator.each do |service_credential_type|
             block.call(service_credential_type)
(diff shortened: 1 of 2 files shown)


lib/topological_inventory/ansible_tower/collector/service_catalog.rb
```diff
@@ -89,6 +98,12 @@ def get_service_instance_nodes(connection)
       def get_service_plan(template)
         template.survey_spec_hash if template.survey_enabled
       end
+
+      private
+
+      def log_external_url(connection_class)
+        logger.info("[EXTERNAL URL] #{tower_hostname}/api/v2/#{connection_class.klass.endpoint}")
```
slemrmartin: You can use `connection.api_url(tower.hostname)` instead of hardcoded version

slemrmartin (changes requested):
I would rather extend these logs: 
- https://github.com/RedHatInsights/topological_inventory-ansible_tower/blob/master/lib/topological_inventory/ansible_tower/collector.rb#L64
- https://github.com/RedHatInsights/topological_inventory-ansible_tower/blob/master/lib/topological_inventory/ansible_tower/collector.rb#L96
- https://github.com/RedHatInsights/topological_inventory-ansible_tower/blob/master/lib/topological_inventory/ansible_tower/collector.rb#L105

by external_url, because the first one is always preceding all these logs without any significant operation (and 2nd/3rd logs are closing the 1st). 
Similarly to all collector repos

agrare:
@slemrmartin since the entity_type != `connection.api.foo` I don't think we'd be able to log more than just the tower base url URL if it was done in the collector_thread.

syncrou (commented):
>maybe I could move this to the base collector class since it's just one method and it could be used across all of the collectors, that way we can just call this moethod in each get_#{resource_type} method call.

☝️ I like this idea.  It seems per @agrare's comment and what I've learned from the current logs that we can't get access to what we actually need to log if we try extending them.

But having the use of a shared method off the base makes a lot of sense to me.

lindgrenj6:
@syncrou https://github.com/RedHatInsights/topological_inventory-providers-common/pull/18

Opened PR on common collector to move the method.

slemrmartin (approved):

slemrmartin:
https://github.com/RedHatInsights/topological_inventory-providers-common/pull/20 
This is a blocker for this PR

syncrou (approved):
LGTM 👍 Happy to get this in.

PR #88, opened 2020-04-24
[WIP] Add heartbeat check for operations worker

lpichler:
Included:

Needed:

- [ ] (https://github.com/RedHatInsights/topological_inventory-providers-common/pull/17)

- added script to check whether heartbeat files are expired and this script will be used in configuration:
 - `bin/heartbeat_check_operations`
 - `bin/heartbeat_check_collector`

Operations:

- heartbeat ticks are in Kafka notification (when we are waiting for message in Kafka)
- in ordering (process - maybe this is not needed it is avoid some delay in network requests)

Collector:
- heartbeat tick is after we fetch data from collections

@miq-bot assign @slemrmartin

### DIFF
--- a/bin/ansible_tower-operations
+++ b/bin/ansible_tower-operations
@@ -1,10 +1,12 @@
 #!/usr/bin/env ruby
 
 $LOAD_PATH << File.expand_path("../lib", __dir__)
+
 require "bundler/setup"
 require "active_support/core_ext/object/blank"
 require "optimist"
 require "topological_inventory/ansible_tower/operations/worker"
+require "topological_inventory/providers/common/heartbeat"
 
 def parse_args
   Optimist.options do
@@ -42,6 +44,15 @@ end
 SourcesApiClient.configure do |config|
   config.scheme = args[:sources_scheme] || "http"
   config.host   = "#{args[:sources_host]}:#{args[:sources_port]}"
+end
+
+ActiveSupport::Notifications.subscribe(/.*heartbeat\.consumer\.kafka.*$/) do |*options|
+  messaging_options = TopologicalInventory::AnsibleTower::Operations::Worker.default_messaging_opts
+  event = ActiveSupport::Notifications::Event.new(*options)
+  if messaging_options[:group_ref] == event.payload[:group_id] && messaging_options[:client_ref] == event.payload[:client_id]
+    heartbeat = TopologicalInventory::Providers::Common::Heartbeat.new("operations-kafka")
+    heartbeat.touch_heartbeat_file
+  end
 end
 
 begin
--- a/bin/heartbeat_check_collector
+++ b/bin/heartbeat_check_collector
@@ -1,0 +1,14 @@
+#!/usr/bin/env ruby
+
+$LOAD_PATH << File.expand_path("../lib", __dir__)
+
+require "bundler/setup"
+require "topological_inventory/providers/common/heartbeat"
+
+HEARTBEATS = %w[collector-standalone collector_pool]
+
+exit_status = HEARTBEATS.any? { |name| TopologicalInventory::Providers::Common::Heartbeat.check(name) }
+
+exit_status = exit_status ? 0 : 1
+
+exit exit_status
--- a/bin/heartbeat_check_operations
+++ b/bin/heartbeat_check_operations
@@ -1,0 +1,14 @@
+#!/usr/bin/env ruby
+
+$LOAD_PATH << File.expand_path("../lib", __dir__)
+
+require "bundler/setup"
+require "topological_inventory/providers/common/heartbeat"
+
+HEARTBEATS = %w[operations operations-kafka]
+
+exit_status = HEARTBEATS.any? { |name| TopologicalInventory::Providers::Common::Heartbeat.check(name) }
+
+exit_status = exit_status ? 0 : 1
+
+exit exit_status
--- a/lib/topological_inventory/ansible_tower/collector.rb
+++ b/lib/topological_inventory/ansible_tower/collector.rb
@@ -3,17 +3,23 @@ require "topological_inventory/providers/common/collector"
 require "topological_inventory/ansible_tower/connection"
 require "topological_inventory/ansible_tower/parser"
 require "topological_inventory/ansible_tower/iterator"
+require "topological_inventory/providers/common/heartbeat"
 
 module TopologicalInventory::AnsibleTower
   class Collector < TopologicalInventory::Providers::Common::Collector
     include Logging
+    include TopologicalInventory::Providers::Common::HeartbeatQueue
 
     require "topological_inventory/ansible_tower/collector/service_catalog"
     include TopologicalInventory::AnsibleTower::Collector::ServiceCatalog
 
     def initialize(source, tower_hostname, tower_user, tower_passwd, metrics,
-                   poll_time: 60, standalone_mode: true)
-      super(source, :poll_time => poll_time, :standalone_mode => standalone_mode)
+                   poll_time: 60, standalone_mode: true, heartbeat_queue: nil)
+      if standalone_mode
+        heartbeat_queue = heartbeat('collector-standalone')
+      end
+
+      super(source, :poll_time => poll_time, :standalone_mode => standalone_mode, :heartbeat_queue => heartbeat_queue)
 
       self.connection_manager = TopologicalInventory::AnsibleTower::Connection.new
       self.tower_hostname     = tower_hostname
@@ -22,15 +28,27 @@ module TopologicalInventory::AnsibleTower
       self.metrics            = metrics
     end
 
+    def perform_collector_threads
+      ensure_collector_threads
+
+      collector_threads.each_value do |thread|
+        thread.join
+      end
+    end
+
     def collect!
       until finished?
-        ensure_collector_threads
+        if standalone_mode
+          heartbeat_queue.run_in_parallel_with do
+            perform_collector_threads
 
-        collector_threads.each_value do |thread|
-          thread.join
+            sleep(poll_time)
+          end
+        else
+          perform_collector_threads
+
+          stop
         end
-
-        standalone_mode ? sleep(poll_time) : stop
       end
     end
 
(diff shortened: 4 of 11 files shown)


lib/topological_inventory/ansible_tower/operations/worker.rb
```diff
@@ -2,12 +2,14 @@
...
 require "topological_inventory/ansible_tower/operations/processor"
 require "topological_inventory/ansible_tower/operations/source"
+require "topological_inventory/providers/common/heartbeat"
 
 module TopologicalInventory
   module AnsibleTower
     module Operations
       class Worker
         include Logging
+        include TopologicalInventory::Providers::Common::HeartbeatQueue
```
slemrmartin, line 12: Is the include there necessary?

bin/heartbeat_check_operations
```diff
@@ -0,0 +1,13 @@
...
+require "bundler/setup"
+require "topological_inventory/providers/common/heartbeat"
+
+exit_status = false
+
+%w[collector-standalone collector-pool].each do |heartbeat_name|
+  exit_status = TopologicalInventory::Providers::Common::Heartbeat.check(heartbeat_name) unless exit_status
+end
+
+exit exit_status
```
slemrmartin, line 14: This script is "ExecAction" of liveness probe, right?. It expects 0 (https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/) as a success, I'm not sure how it handles boolean, please convert

bin/heartbeat_check_operations
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env ruby
```
slemrmartin, line 1: File names are switched (operations for collectors, collectors for operations)

lib/topological_inventory/ansible_tower/operations/core/service_order_mixin.rb
```diff
@@ -16,18 +16,26 @@ def order
...
 
+            heartbeat('operations').touch_heartbeat_file
+
             job_type = parse_svc_offering_type(service_offering)
 
             logger.info("ServiceOffering#order: Task(id: #{task_id}): Ordering ServiceOffering(id: #{service_offering.id}, source_ref: #{service_offering.source_ref}): Launching Job...")
             job = client.order_service(job_type, service_offering.source_ref, order_params)
             logger.info("ServiceOffering#order: Task(id: #{task_id}): Ordering ServiceOffering(id: #{service_offering.id}, source_ref: #{service_offering.source_ref}): Job(:id #{job&.id}) has launched.")
 
+            heartbeat('operations').touch_heartbeat_file
```
slemrmartin: as we discussed, it'll be better to do "ticks" in separate thread with own timeout during processing a operation ( bigger than tower connection's timeout  - https://github.com/RedHatInsights/topological_inventory-ansible_tower/blob/8e3b4a6e777b5a24e8e28ad5a0aa2a63900534b9/lib/topological_inventory/ansible_tower/connection.rb#L8)

lib/topological_inventory/ansible_tower/operations/core/service_order_mixin.rb
```diff
@@ -16,18 +16,26 @@ def order
             update_task(task_id, :state => "running", :status => "ok")
 
             service_plan          = topology_api_client.show_service_plan(service_plan_id.to_s) if service_plan_id
+            heartbeat('operations').touch_heartbeat_file
```
slemrmartin: Notifications in operations file have 'operations-kafka' name. Shouldn't it touch the same file?

bin/ansible_tower-operations
```diff
@@ -44,6 +45,15 @@ SourcesApiClient.configure do |config|
   config.host   = "#{args[:sources_host]}:#{args[:sources_port]}"
 end
 
+ActiveSupport::Notifications.subscribe(/.*heartbeat\.consumer\.kafka.*$/) do |*options|
```
slemrmartin, line 49: If the Kafka notifications is emited once per 10 seconds and heartbeat timeout is also 10 seconds, it can cause problems in edge cases

slemrmartin (changes requested):
The idea is good, just couple of changes are needed

bin/ansible_tower-operations
lpichler: I increased the limit for checking heartbeat file to 12 seconds so it should be enough

lib/topological_inventory/ansible_tower/operations/worker.rb
lpichler: yes it is

lib/topological_inventory/ansible_tower/operations/core/service_order_mixin.rb
lpichler: I rather using two files and in check heartbeatfile beat I am considering both.

lpichler:
@slemrmartin updated
tested also with collector.
heartbeat mechanism waits for ansible connections max 65 second (default) as requests from ansible has timeout 60 seconds.

slemrmartin:
@lindgrenj6 please review

lib/topological_inventory/ansible_tower/operations/worker.rb
slemrmartin: ok, now it is :D

lib/topological_inventory/ansible_tower/operations/core/ansible_tower_client.rb
```diff
@@ -2,6 +2,7 @@
 require "topological_inventory/ansible_tower/connection"
 require "topological_inventory/ansible_tower/operations/core/sources_api_client"
 require "topological_inventory/ansible_tower/operations/core/topology_api_client"
+require "topological_inventory/providers/common/heartbeat"
```
slemrmartin: shouldn't be here, heartbeat is not used by the tower client object

lib/topological_inventory/ansible_tower/operations/service_offering.rb
```diff
@@ -9,6 +9,7 @@ module AnsibleTower
     module Operations
       class ServiceOffering
         include Logging
+        include TopologicalInventory::Providers::Common::HeartbeatQueue
```
slemrmartin: it's no longer used by `ServiceOrderMixin`

lib/topological_inventory/ansible_tower/operations/service_plan.rb
```diff
@@ -8,6 +8,7 @@ module AnsibleTower
     module Operations
       class ServicePlan
         include Logging
+        include TopologicalInventory::Providers::Common::HeartbeatQueue
```
slemrmartin: it's no longer used by `ServiceOrderMixin`

lib/topological_inventory/ansible_tower/collector.rb
```diff
@@ -68,8 +86,15 @@ def collector_thread(connection, entity_type)
       sweep_scope = Set.new
       cnt = 0
       refresh_state_part_collected_at = nil
-      # each on ansible_tower_client's enumeration makes pagination requests by itself
-      send("get_#{entity_type}", connection).each do |entity|
+
+      heartbeat_thread_method = standalone_mode ? :run_in_parallel_with : :run_thread_queue_in_parallel_with
+      entities = heartbeat_queue.send(heartbeat_thread_method) do
+                   send("get_#{entity_type}", connection)
```
slemrmartin, line 92: what's the real timeout there? Is it open_timeout + read_timeout or the read_timeout includes the open_timeout?
Just thinking that the heartbeat timeout is set to 65s and open_timeout + (default) read_timeout is 70s

slemrmartin (commented):
Few minor comments, but otherwise awesome :+1:

bin/heartbeat_check_collector
```diff
@@ -0,0 +1,16 @@
...
+$LOAD_PATH << File.expand_path("../lib", __dir__)
+
+require "bundler/setup"
+require "topological_inventory/providers/common/heartbeat"
+
+exit_status = false
+
+%w[collector-standalone collector_pool].each do |heartbeat_name|
+  exit_status = TopologicalInventory::Providers::Common::Heartbeat.check(heartbeat_name) unless exit_status
+end
```
lindgrenj6: This section could be changed to:
```
HEARTBEATS = %w[collector-standalone collector_pool]

exit_status = HEARTBEATS.any? { |name| TopologicalInventory::Providers::Common::Heartbeat.check(heartbeat_name) }
```

I think the 2 heart beats should be a constant, any `.any?` does exactly what you need with exit_status.

bin/heartbeat_check_operations
```diff
@@ -0,0 +1,16 @@
...
+
+require "bundler/setup"
+require "topological_inventory/providers/common/heartbeat"
+
+exit_status = false
+
+%w[operations operations-kafka].each do |heartbeat_name|
+  exit_status = TopologicalInventory::Providers::Common::Heartbeat.check(heartbeat_name) unless exit_status
+end
+
```
lindgrenj6: Same as above

lib/topological_inventory/ansible_tower/collector.rb
lpichler: it will 65 seconds keep touching heartbeat file(each 10 seconds.) so last time when file was touched is on 60th seconds, and heartbeat check time out is set to 12 seconds.
so basically heartbeat check will start to fail(return value other than 0) after 72seconds when request started.

According to my testing read_timeout includes the open_timeout.

Hopefully I answered your question. :)

lpichler:
@slemrmartin @lindgrenj6 changes are done, thanks for your suggestions 👍 


now I need to change version of `topological_inventory-providers-common` gem with this PR https://github.com/RedHatInsights/topological_inventory-providers-common/pull/17 .

lindgrenj6 (commented):
This LGTM, just waiting on the providers-common PR to go in and get released.

bin/ansible_tower-operations
```diff
@@ -44,6 +46,15 @@ SourcesApiClient.configure do |config|
...
+  messaging_options = TopologicalInventory::AnsibleTower::Operations::Worker.default_messaging_opts
+  event = ActiveSupport::Notifications::Event.new(*options)
+  if messaging_options[:group_ref] == event.payload[:group_id] && messaging_options[:client_ref] == event.payload[:client_id]
+    heartbeat = TopologicalInventory::Providers::Common::Heartbeat.new("operations-kafka")
+    heartbeat.touch_heartbeat_file
```
roliveri, line 54: This is just a suggestion. For patterns like this, it may be cleaner to have a class method of the `Heartbeat` class memoize and cache the instance and perform the operation. For example:
```
TopologicalInventory::Providers::Common::Heartbeat.touch_heartbeat_file("operations-kafka")
```
Here, the `touch_heartbeat_file` class method would instantiate the instance and hash it based on name, if needed. Then perform the operation on the hashed value.

This comment applies more to https://github.com/RedHatInsights/topological_inventory-providers-common/pull/17, but the use case is more visible here.

Here again, this is only a suggestion, as this is a cosmetic issue.

lpichler:
I am closing this as there is replacement https://github.com/RedHatInsights/topological_inventory-providers-common/pull/48 -  this PR could be used later if needed.

thanks!

PR #90, opened 2020-05-11
Add Source UID to log statements

lindgrenj6:
https://projects.engineering.redhat.com/browse/TPINVTRY-935

This shows which source is being refreshed in the tower collector. 

cc @syncrou @gtanzillo @slemrmartin @lahavyuv86

### DIFF
--- a/lib/topological_inventory/ansible_tower/collector.rb
+++ b/lib/topological_inventory/ansible_tower/collector.rb
@@ -61,7 +61,7 @@ module TopologicalInventory::AnsibleTower
     # Thread's main for collecting one entity type's data
     def collector_thread(connection, entity_type)
       refresh_state_uuid, refresh_state_started_at = SecureRandom.uuid, Time.now.utc
-      logger.info("[START] Collecting #{entity_type} with :refresh_state_uuid => '#{refresh_state_uuid}'")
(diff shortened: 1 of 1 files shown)


lib/topological_inventory/ansible_tower/collector.rb
```diff
@@ -93,20 +93,24 @@ def collector_thread(connection, entity_type)
...
-      logger.error("Error collecting: #{entity_type} :refresh_state_uuid => #{refresh_state_uuid}, message => #{e.message} #{e.backtrace.join("\n")}")
+      logger.error("Error collecting: #{entity_type}, :source_id => #{source_id}, :refresh_state_uuid => #{refresh_state_uuid}, message => #{e.message} #{e.backtrace.join("\n")}")
     end
 
     def inventory_name
       "AnsibleTower"
     end
+
+    def source_id
+      @source_id ||= Settings.sources.select { |s| s.source == source }&.first&.source_id
```
slemrmartin: @lindgrenj6 Collector doesn't initialize Config thus it doesn't know Settings so it can't work in standalone mode.

lindgrenj6: Thanks for catching that, I updated it to log the Source UID since we'll always have that field.

slemrmartin (approved):
:+1: LGTM

PR #94, opened 2020-05-28
Support for Tower 3.7

slemrmartin:
Survey choices are Array instead of String since 3.6/3.7? 
Was a Tower 3.5 bug, because it should be always array according to 3.5 doc:

- https://docs.ansible.com/ansible-tower/3.5.0/html/towerapi/api_ref.html#/Job_Templates/Job_Templates_job_templates_survey_spec_list

Some Inventory fields are deprecated, some of them in 3.6,  others in 3.7
Deprecation announced in Release notes 3.5.0

- https://docs.ansible.com/ansible-tower/3.6.0/html/towerapi/api_ref.html#/Inventories/Inventories_inventories_list
- https://docs.ansible.com/ansible-tower/latest/html/towerapi/api_ref.html#/Inventories/Inventories_inventories_list

---

[TPINVTRY-945](https://projects.engineering.redhat.com/browse/TPINVTRY-945)


slemrmartin:
### Ansible Tower Release notes

3.7.0 Added a new field to jobs for tracking the date/time a job is canceled 
- **enhancement, can be added later**

3.6.3 Deprecated the CloudForms inventory source 
- **not related**

3.6.0 Added the ability to create approval nodes in workflow job templates 
- don't know what it is - **can this influence Approval inventories task?**

3.6.0 Removed the deprecated ask_ fields in the Job Details view to reduce confusion 
- **we don't collect it**

3.6.0 Removed the following API endpoints:
- Job template credential field (in either v1 or v2)
- - we're using JobTemplate.summary_fields.credentials
- - **needs to be tested against 3.6+**
- Corresponding credential field on workflow job template nodes
- - **needs to be tested against 3.6+**
- **we can temporarily skip collecting of credentials as Catalog doesn't need it now**

syncrou (approved):
This looks good. The only question I initially had was should we be allowing for backward compatibility on the deprecated attributes for service_inventory.  However, looking at the Tower API docs and our GA document around supported Tower versions, it makes sense to start at v3.6 and to remove the deprecated attributes like what is done here. I see no need to provide for the v3.5 attributes since Tower v3.5 was never listed as a supported version.

slemrmartin:
@syncrou the question is if Catalog is using these attributes now?

gmcculloug (approved):
@slemrmartin Thanks for the quick turnaround on this issue.  👍

lindgrenj6 (approved):
LGTM, initially had the same question as Drew about leaving those in and using the `&.` operator to safely pull them in, but if they're already deprecated as of 3.6 why even collect them since they went away in the next version.

slemrmartin:
@lindgrenj6 `&.` operator raises exception if caller exists and doesn't respond. 
`.try` is working for this case (but is significantly slower)

lindgrenj6:
@slemrmartin Just to be safe I double checked Catalog's code (just grep'd for each name) - no mention of those fields, so we should be good.

PR #95, opened 2020-05-28
Update Endpoint's availability status

slemrmartin:
**Issue**: https://github.com/RedHatInsights/topological_inventory-providers-common/issues/24

Availability check updates `Source.availability_status` and `Endpoint.availability_status` + `Endpoint.availability_status_error`

Based on the latest implementation of availabity_checker in Satellite operations:
- https://github.com/RedHatInsights/topological_inventory-satellite/pull/14

Structure of class changed as close as possible to the satellite repo. 
Other repos are basically copy-paste so the next-step will be moving it to providers-common (https://github.com/RedHatInsights/topological_inventory-providers-common/pull/25)

---

- [x] **depends on** https://github.com/RedHatInsights/sources-api-client-ruby/pull/19
- [x] **depends on** https://github.com/RedHatInsights/sources-api-client-ruby/pull/20
- [x] **depends on** https://github.com/RedHatInsights/topological_inventory-providers-common/pull/27

---

[TPINVTRY-948](https://projects.engineering.redhat.com/browse/TPINVTRY-948)


lib/topological_inventory/ansible_tower/operations/source.rb
```diff
@@ -10,58 +10,113 @@ class Source
...
+          end
 
-          auth_id = endpoint_authentications.first.id
-          auth = Core::AuthenticationRetriever.new(auth_id, identity).process
-          return STATUS_UNAVAILABLE unless auth
+          is_missing
+        end
+
+        def connection_status
+          return [STATUS_UNAVAILABLE, ERROR_MESSAGES[:endpoint]] unless endpoint
```
gtanzillo: I think this should be `ERROR_MESSAGES[:endpoint_not_found]` instead of `ERROR_MESSAGES[:endpoint]`

slemrmartin:
@syncrou the `availablilty_status_error` is not `'nullable': true` in Sources API so this code can't delete the error message when source becomes available. (https://github.com/RedHatInsights/sources-api/pull/227 ?)

lib/topological_inventory/ansible_tower/operations/source.rb
```diff
@@ -10,58 +10,113 @@ class Source
...
+        end
 
+        def connection_check
           connection = ::TopologicalInventory::AnsibleTower::Connection.new
-          connection = connection.connect(endpoint.host, auth.username, auth.password)
+          connection = connection.connect(endpoint.host, authentication.username, authentication.password)
           connection.api.version
 
-          STATUS_AVAILABLE
+          [STATUS_AVAILABLE, nil]
```
lindgrenj6, line 66: Small nit, but this can stay as just `STATUS_AVAILABLE`, had to try it myself to make sure but heres what I found:
```ruby
[ ⚛ ] ~ $ irb
Welcome to IRB. You are using ruby 2.6.6p146 (2020-03-31 revision 67876) [x86_64-linux]. Have fun ;)
>> status, err = [:status, :error]
>> status #=> :status
>> err #=> :error
>> status, err = :status
>> status #=> :status
>> err #=> nil
>> 

```

EDIT: 
Although that would be a bit harder to read. I'm ok with either, explicit might be better than implicit in this case.

lib/topological_inventory/ansible_tower/operations/source.rb
```diff
@@ -10,58 +10,113 @@ class Source
...
+        def update_endpoint(status, error_message)
+          if endpoint.nil?
+            logger.error("Source#availability_check - Failed to update Endpoint for Source id:#{source_id}. Endpoint not found")
+            return
+          end
+
+          endpoint_update = SourcesApiClient::Endpoint.new
+
+          endpoint_update.availability_status = status
+          endpoint_update.availability_status_error = error_message # if error_message
```
lindgrenj6: does this `# if error_message` need to be uncommented? maybe we could just default it to `nil` that way we wouldn't have to check it. 

EDIT:
After further review it looks like it _is_ defaulting to nil! Maybe we can just remove the comment then?

lib/topological_inventory/ansible_tower/operations/source.rb
```diff
@@ -10,58 +10,113 @@ class Source
...
+          @endpoint ||= api_client.list_source_endpoints(source_id)&.data&.detect(&:default)
+        end
+
+        def authentication
+          return @authentication if @authentication.present?
+
+          endpoint_authentications = api_client.list_endpoint_authentications(endpoint.id.to_s).data || []
+          return nil if endpoint_authentications.empty?
+
+          auth_id = endpoint_authentications.first.id
```
lindgrenj6, line 114: I'm just more curious here - it looks like we have a few endpoints that have more than one authentication, found with this query: `Endpoint.all.map(&:authentications).map(&:count).uniq`.

Should we make this loop through the authentications if one fails or is that basically never going to happen? I know it would be difficult, but maybe we could memoize `endpoint_authentications` and if one fails, just change `@authentication` to the next one and `retry`. 

I'm just shooting in the dark here, I'm still unfamiliar with a lot of this but just thought I'd put out my thoughts .

spec/operations/source_spec.rb
```diff
@@ -2,33 +2,158 @@
...
+      {
+        'params' => {
+          'source_id'       => source_id,
+          'external_tenant' => external_tenant,
+          'timestamp'       => Time.now.utc
         }
+      }
+    end
+
+    let(:list_endpoints_response) { "{\"meta\":{\"count\":1,\"limit\":100,\"offset\":0},\"links\":{\"first\":\"#{sources_api_path}/sources/#{source_id}/endpoints?offset=0\",\"last\":\"#{sources_api_path}/sources/#{source_id}/endpoints?offset=0\"},\"data\":[{\"default\":true,\"host\":\"10.0.0.1\",\"id\":\"#{endpoint_id}\",\"path\":\"/\",\"role\":\"ansible\",\"scheme\":\"https\",\"source_id\":\"#{source_id}\",\"tenant\":\"#{external_tenant}\"}]}"}
```
lindgrenj6: I think on these we can remove the `meta` fields - the client will still parse the object fine and it makes it easier to read since we're only inserting the fields we need in the response.

lib/topological_inventory/ansible_tower/operations/source.rb
slemrmartin: me too, anyway it's interesting, didn't know about that

lib/topological_inventory/ansible_tower/operations/source.rb
slemrmartin: @lindgrenj6 that's surprise for me! @syncrou do you know what does it mean? If there are more authentications for one endpoint, which one should be used for operations and collectors?

slemrmartin: anyway this problem is not related only to tower operations, so I would recommend to merge this PR with this and solve it either in following PR in providers-common, or make a separate issue because it's related to all operations and orchestrator

lindgrenj6: Yes, this is most likely an edge case where other sources have multiple auth's for one endpoint, it shouldn't happen with anything we collect.

I'll create an issue on the `sources-api` so we're at least aware of it, it should be fine as far as the collectors goes.

syncrou: It appears there was discussion around handling multiple Authentications per Endpoint. So the modeling is there.  Though there is not currently a way to designate a `current` or `default` authentication.

For this PR `.first` will suffice as it appears the ability is only exposed via the API and not in the UI.

lindgrenj6 (approved):

lib/topological_inventory/ansible_tower/operations/core/authentication_retriever.rb
```diff
@@ -35,7 +35,9 @@ def process
               :headers => headers
             }
             response = RestClient::Request.new(request_options).execute
-            ::SourcesApiClient::Authentication.new(JSON.parse(response.body))
+            params = JSON.parse(response.body)
+            params.delete('tenant')
```
syncrou, line 39: @slemrmartin why do we need to delete the `tenant` param here?

lib/topological_inventory/ansible_tower/operations/source.rb
```diff
@@ -10,58 +10,113 @@ class Source
...
+          :endpoint_not_found       => "Endpoint not found in Sources API",
+        }.freeze
+
+        attr_accessor :params, :request_context, :source_id
 
         def initialize(params = {}, request_context = nil)
-          @params  = params
-          @context = request_context
-          @source_id = nil
+          self.params          = params
```
syncrou, line 21: Was wondering why we need to preface with `self.`?

lib/topological_inventory/ansible_tower/operations/core/authentication_retriever.rb
slemrmartin: because the public API doesn't return tenant (https://github.com/RedHatInsights/sources-api/pull/185) but internal API does. And the client is generated from public API

lib/topological_inventory/ansible_tower/operations/source.rb
slemrmartin: I just want to be consistent with other classes/repos in topological inventory

syncrou (approved):
LGTM 👍

PR #96, opened 2020-05-28
Availability check: "last_checked_at" and "last_available_at" update

slemrmartin:
**Issue** https://github.com/RedHatInsights/sources-api/issues/206

Availability check checks the `Endpoint.last_checked_at` and skips new testing if last check was less than 5 minutes ago. 
This will prevent processing many availability checks with big kafka lag. 

---

- [x] **depends on** #95 (needs rebase after merge)

---

[TPINVTRY-944](https://projects.engineering.redhat.com/browse/TPINVTRY-944)

### DIFF
--- a/lib/topological_inventory/ansible_tower/operations/source.rb
+++ b/lib/topological_inventory/ansible_tower/operations/source.rb
@@ -15,6 +15,8 @@ module TopologicalInventory
           :endpoint_not_found       => "Endpoint not found in Sources API",
         }.freeze
 
+        LAST_CHECKED_AT_THRESHOLD = 5 # minutes
+
         attr_accessor :params, :request_context, :source_id
 
         def initialize(params = {}, request_context = nil)
@@ -25,6 +27,8 @@ module TopologicalInventory
 
         def availability_check
           return if params_missing?
+
+          return if checked_recently?
 
           status, error_message = connection_status
 
@@ -43,12 +47,21 @@ module TopologicalInventory
           is_missing = false
           required_params.each do |attr|
             if (is_missing = params[attr].blank?)
-              logger.error("Source#availability_check - Missing #{attr} for the availability_check request")
+              logger.error("Source#availability_check - Missing #{attr} for the availability_check request [Source ID: #{source_id}]")
               break
             end
           end
 
           is_missing
+        end
+
+        def checked_recently?
+          return false if endpoint.nil?
+
+          checked_recently = endpoint.last_checked_at.present? && endpoint.last_checked_at >= LAST_CHECKED_AT_THRESHOLD.minutes.ago
+          logger.info("Source#availability_check - Skipping, last check at #{endpoint.last_checked_at} [Source ID: #{source_id}] ") if checked_recently
+
+          checked_recently
         end
 
         def connection_status
@@ -59,13 +72,14 @@ module TopologicalInventory
         end
 
         def connection_check
+          check_time
           connection = ::TopologicalInventory::AnsibleTower::Connection.new
           connection = connection.connect(endpoint.host, authentication.username, authentication.password)
           connection.api.version
 
           [STATUS_AVAILABLE, nil]
         rescue => e
-          logger.error("Failed to connect to Source id:#{source_id} - #{e.message}")
+          logger.error("Source#availability_check - Failed to connect to Source id:#{source_id} - #{e.message}")
           [STATUS_UNAVAILABLE, e.message]
         end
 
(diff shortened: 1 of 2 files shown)


lib/topological_inventory/ansible_tower/operations/source.rb
```diff
@@ -15,6 +15,8 @@ class Source
           :endpoint_not_found       => "Endpoint not found in Sources API",
         }.freeze
 
+        LAST_CHECKED_AT_THRESHOLD = 5 # minutes
```
syncrou: If we want this to be in minutes - potentially just declare `5.minutes`.

lib/topological_inventory/ansible_tower/operations/source.rb
```diff
@@ -43,14 +47,23 @@ def params_missing?
...
+              logger.error("Source#availability_check - Missing #{attr} for the availability_check request [Source ID: #{source_id}]")
               break
             end
           end
 
           is_missing
         end
 
+        def checked_recently?
+          return false if endpoint.nil?
```
syncrou, line 59: On prod currently we have sources without an `Endpoint`.  They have `Application` records.

How would we know if the `source.applications.first.last_checked_at` was `checked_recently?`

Adam's comment was saying the same I believe (https://github.com/RedHatInsights/sources-api/issues/206#issue-587773644)

Though I see in the PR description, this was only to add the additional updates to `Endpoint` so we could make the additional changes with follow up PRs

lib/topological_inventory/ansible_tower/operations/source.rb
```diff
@@ -59,13 +72,14 @@ def connection_status
         end
 
         def connection_check
+          check_time
```
syncrou, line 75: Is this here to preload `@check_time`?

lib/topological_inventory/ansible_tower/operations/source.rb
```diff
@@ -93,8 +109,10 @@ def update_endpoint(status, error_message)
 
           endpoint_update = SourcesApiClient::Endpoint.new
```
syncrou, line 110: Potentially move `endpoint_update` into its own method? `update_endpoint`?  Would allow us to test easier and be more consistent.

lib/topological_inventory/ansible_tower/operations/source.rb
```diff
@@ -26,6 +28,8 @@ def initialize(params = {}, request_context = nil)
         def availability_check
           return if params_missing?
 
+          return if checked_recently?
```
lindgrenj6, line 31: Does this implicitly return true? I assume it does.

lindgrenj6 (commented):
Just one small clarification, otherwise it LGTM. This should satisfy the sources requirement.

lib/topological_inventory/ansible_tower/operations/source.rb
gmcculloug: @lindgrenj6 The `return` without a parameter will pass a `nil` back to the caller.  Is that what you are referring to?

lib/topological_inventory/ansible_tower/operations/source.rb
slemrmartin: Actually this is an `update_endpoint` method :)

lib/topological_inventory/ansible_tower/operations/source.rb
slemrmartin: yes, to ensure the time is grabbed before checking and to be the same for both timestamps

lib/topological_inventory/ansible_tower/operations/source.rb
slemrmartin: We are not able to do availability check for sources without endpoint now. 
Supporting it should be new feature (and better to solve it as PR following  https://github.com/RedHatInsights/topological_inventory-providers-common/pull/25)

lib/topological_inventory/ansible_tower/operations/source.rb
slemrmartin: the `checked_recently?` never returns nil, this line returns nil and caller doesn't care about return value of operation

lib/topological_inventory/ansible_tower/operations/source.rb
slemrmartin: I guess availability check of applications is not a responsibility of our operation, should be checked externally: https://github.com/RedHatInsights/sources-api/blob/ab24b6c710c8d31eadf4f5068a01d38863b8c022/app/controllers/api/v1/sources_controller.rb#L56-L55

lib/topological_inventory/ansible_tower/operations/source.rb
syncrou: haha! It is.  I didn't see that the line numbers were skipped.

syncrou (approved):
LGTM 👍

lindgrenj6 (approved):

PR #98, opened 2020-06-02
Remove collection of `insights_credential` on service_inventory

lindgrenj6:
https://projects.engineering.redhat.com/browse/TPINVTRY-951

This will work (for now) to get past this - I can't find a good way to reproduce this for now, so this will get us past it for the time being. 

cc @syncrou @slemrmartin

### DIFF
--- a/lib/topological_inventory/ansible_tower/parser/service_inventory.rb
+++ b/lib/topological_inventory/ansible_tower/parser/service_inventory.rb
@@ -15,7 +15,6 @@ module TopologicalInventory::AnsibleTower
               "variables"                       => inventory.variables,
               "total_inventory_sources"         => inventory.total_inventory_sources,
               "inventory_sources_with_failures" => inventory.inventory_sources_with_failures,
-              "insights_credential"             => inventory.insights_credential,
               "pending_deletion"                => inventory.pending_deletion,
             }
           )


lib/topological_inventory/ansible_tower/parser/service_inventory.rb
```diff
@@ -15,7 +15,7 @@ def parse_service_inventory(inventory)
               "variables"                       => inventory.variables,
               "total_inventory_sources"         => inventory.total_inventory_sources,
               "inventory_sources_with_failures" => inventory.inventory_sources_with_failures,
-              "insights_credential"             => inventory.insights_credential,
+              "insights_credential"             => inventory&.insights_credential,
```
slemrmartin: it won't work, because the inventory exists, but method doesn't.
Try `insights_credential_id`, it probably exists.

Or I'm for deleting the whole line, I doubt we are using relation inside jsonb column somewhere.

slemrmartin (changes requested):

lib/topological_inventory/ansible_tower/parser/service_inventory.rb
lindgrenj6: good point, I'm pretty sure we aren't using it anywhere either (and why would we? thats what sources is for right).

Pushed up a change removing the line.

slemrmartin (approved):
:+1: LGTM

PR #99, opened 2020-06-05
Targeted refresh worker

slemrmartin:
Targeted refresh of a job/workflow job is similar to order operation - it starts listening on Kafka queue.

- It can handle refresh of multiple jobs/workflows with the same Source.
- It requests Jobs, in case of Workflow its workflow nodes and jobs associated with the nodes.
Doesn't collect workflow nodes of child workflows (as we discussed)
- It also doesn't collect credentials for now
- All other relations are collected by full refresh so it's not needed to collect, refresh is creating `lazy_links` which are assigned to appropriate record in persister
- ~~removes** collecting of jobs/workflow job nodes from full refresh~~
- Refreshes are scheduled by https://github.com/slemrmartin/topological_inventory-scheduler/blob/master/lib/topological_inventory/scheduler/worker.rb (new service)

---

**Depends on**
- [x] Kafka topic `platform.topological-inventory.collector-ansible-tower` needs to be registered on CI/QA/Prod*
- [x] https://github.com/RedHatInsights/topological_inventory-ansible_tower/pull/117

Kafka Topic added to Prod [here](https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/8169)

---

**Dependents**
- https://github.com/slemrmartin/topological_inventory-scheduler
- https://github.com/RedHatInsights/e2e-deploy/pull/1868
---

[TPINVTRY-678](https://projects.engineering.redhat.com/browse/TPINVTRY-678)

### DIFF
--- a/lib/topological_inventory/ansible_tower/collector/service_catalog.rb
+++ b/lib/topological_inventory/ansible_tower/collector/service_catalog.rb
@@ -60,9 +60,9 @@ module TopologicalInventory::AnsibleTower
         end
       end
 
-      def get_service_instance_nodes(connection, query_params, on_premise: false, receptor_receiver: nil, receptor_params: {})
+      def get_service_instance_nodes(connection, query_params, workflow: nil, on_premise: false, receptor_receiver: nil, receptor_params: {})
         tower_types = %i[workflow_job_nodes]
-        get_tower_objects(connection, query_params, tower_types, :on_premise => on_premise, :receptor_receiver => receptor_receiver, :receptor_params => receptor_params) do
+        get_tower_objects(connection, query_params, tower_types, :slug => workflow&.related&.workflow_nodes, :on_premise => on_premise, :receptor_receiver => receptor_receiver, :receptor_params => receptor_params) do
           # transformation of Workflow Job Node to parser-compatible hash
           # + subqueries for node's credentials
           lambda do |service_instance_node|
@@ -85,14 +85,15 @@ module TopologicalInventory::AnsibleTower
       # @param connection [AnsibleTowerClient::Connection | TopologicalInventory::AnsibleTower::Receptor::ApiClient]
       # @param query_params [Hash] API query params
       # @param tower_types [Array<Symbol>] i.e. %i[job_templates workflow_job_templates]
+      # @param slug [String] optional, slug for calling "find_all_by_url" instead of "all"
       # @param on_premise [Boolean] Ansible Tower placement (in public/on premise)
       # @param receptor_receiver [TopologicalInventory::AnsibleTower::Receptor::AsyncReceiver] Receiver for receptor's asynchronous responses
       # @param receptor_params [Hash] Params for Receptor node's Catalog HTTP plugin (@see https://github.com/mkanoor/receptor-catalog/README.md)
-      def get_tower_objects(connection, query_params, tower_types, on_premise: false, receptor_receiver: nil, receptor_params: {})
+      def get_tower_objects(connection, query_params, tower_types, slug: nil, on_premise: false, receptor_receiver: nil, receptor_params: {})
         api_calls_block = lambda do |&block|
           #
           # Initializing API Calls
-          api_objects = init_api_objects(connection, on_premise, query_params, receptor_params, receptor_receiver, tower_types)
+          api_objects = init_api_objects(connection, on_premise, query_params, receptor_params, receptor_receiver, tower_types, slug)
           #
           # Getting custom lambda block for transformation of API object to Parser compatible object
           parsing_transformation = yield if block_given?
@@ -120,20 +121,28 @@ module TopologicalInventory::AnsibleTower
         end
       end
 
-      def init_api_objects(connection, on_premise, query_params, receptor_params, receptor_receiver, tower_types)
+      def init_api_objects(connection, on_premise, query_params, receptor_params, receptor_receiver, tower_types, slug = nil)
         tower_types.collect do |entity_type|
           #
           # Creating Ansible/Receptor API client objects
           api_obj = on_premise ? connection.api.send(entity_type, receptor_receiver) : connection.api.send(entity_type)
           #
           # Logging Tower Full path
-          log_external_url("#{connection_manager.api_url(tower_hostname)}/#{api_obj.klass.endpoint}")
+          log_external_url("#{connection_manager.api_url(tower_hostname)}/#{slug || api_obj.klass.endpoint}")
           #
           # Calling Tower API
-          if on_premise
-            api_obj.all(query_params, receptor_params)
+          if slug
+            if on_premise
+              api_obj.find_all_by_url(slug, query_params, receptor_params)
+            else
+              api_obj.find_all_by_url(slug, query_params)
+            end
           else
-            api_obj.all(query_params)
+            if on_premise
+              api_obj.all(query_params, receptor_params)
+            else
+              api_obj.all(query_params)
+            end
           end
         end
       end
--- a/lib/topological_inventory/ansible_tower/messaging_client.rb
+++ b/lib/topological_inventory/ansible_tower/messaging_client.rb
@@ -1,0 +1,92 @@
+require "manageiq-messaging"
+
+module TopologicalInventory
+  module AnsibleTower
+    class MessagingClient
+      OPERATIONS_QUEUE_NAME  = "platform.topological-inventory.operations-ansible-tower".freeze
+      JOB_REFRESH_QUEUE_NAME = "platform.topological-inventory.collector-ansible-tower".freeze
+
+      # Kafka host name
+      attr_accessor :queue_host
+      # Kafka port
+      attr_accessor :queue_port
+
+      def initialize
+        @queue_host = 'localhost'
+        @queue_port = 9092
+      end
+
+      def self.default
+        @@default ||= new
+      end
+
+      def configure
+        yield(self) if block_given?
+      end
+
+      # Instance of messaging client for Worker
+      def worker_listener
+        @worker_listener ||= ManageIQ::Messaging::Client.open(worker_listener_opts)
+      end
+
+      # Instance of messaging client for Service Order
+      def job_refresh_publisher
+        @job_refresh_publisher ||= ManageIQ::Messaging::Client.open(job_refresh_publisher_opts)
+      end
+
+      def targeted_refresh_listener
+        @targeted_refresh_listener ||= ManageIQ::Messaging::Client.open(targeted_refresh_listener_opts)
+      end
+
+      def worker_listener_queue_opts
+        {
+          :auto_ack    => false,
+          :max_bytes   => 50_000,
+          :service     => OPERATIONS_QUEUE_NAME,
+          :persist_ref => "topological-inventory-operations-ansible-tower"
+        }
+      end
+
+      def targeted_refresh_listener_queue_opts
+        {
+          :service     => JOB_REFRESH_QUEUE_NAME,
+          :persist_ref => "topological-inventory-collector-ansible-tower"
+        }
+      end
+
+      private
+
+      def worker_listener_opts
+        {
+          :client_ref => default_client_ref,
+          :host       => @queue_host,
+          :port       => @queue_port,
+          :protocol   => :Kafka
+        }
+      end
+
+      def targeted_refresh_listener_opts
+        {
+          :client_ref => default_client_ref,
+          :encoding   => 'json',
+          :host       => @queue_host,
+          :port       => @queue_port,
+          :protocol   => :Kafka
+        }
+      end
+
+      def job_refresh_publisher_opts
+        {
+          :encoding => 'json',
+          :host     => @queue_host,
+          :port     => @queue_port,
+          :protocol => :Kafka,
+        }
+      end
+
+      def default_client_ref
+        ENV['HOSTNAME'].presence || SecureRandom.hex(4)
+      end
+    end
+  end
+end
--- a/lib/topological_inventory/ansible_tower/messaging_client.rb
+++ b/lib/topological_inventory/ansible_tower/messaging_client.rb
@@ -4,7 +4,7 @@ module TopologicalInventory
   module AnsibleTower
     class MessagingClient
       OPERATIONS_QUEUE_NAME  = "platform.topological-inventory.operations-ansible-tower".freeze
-      JOB_REFRESH_QUEUE_NAME = "platform.topological-inventory.collector-ansible-tower".freeze
+      REFRESH_QUEUE_NAME = "platform.topological-inventory.collector-ansible-tower".freeze
 
       # Kafka host name
       attr_accessor :queue_host
@@ -20,18 +20,17 @@ module TopologicalInventory
         @@default ||= new
       end
 
-      def configure
-        yield(self) if block_given?
+      def self.configure
+        if block_given?
+          yield(default)
+        else
+          default
+        end
       end
 
       # Instance of messaging client for Worker
       def worker_listener
         @worker_listener ||= ManageIQ::Messaging::Client.open(worker_listener_opts)
-      end
-
-      # Instance of messaging client for Service Order
-      def job_refresh_publisher
-        @job_refresh_publisher ||= ManageIQ::Messaging::Client.open(job_refresh_publisher_opts)
       end
 
       def targeted_refresh_listener
@@ -49,7 +48,7 @@ module TopologicalInventory
 
       def targeted_refresh_listener_queue_opts
         {
-          :service     => JOB_REFRESH_QUEUE_NAME,
+          :service     => REFRESH_QUEUE_NAME,
           :persist_ref => "topological-inventory-collector-ansible-tower"
         }
       end
@@ -72,15 +71,6 @@ module TopologicalInventory
           :host       => @queue_host,
           :port       => @queue_port,
           :protocol   => :Kafka
-        }
-      end
-
-      def job_refresh_publisher_opts
-        {
-          :encoding => 'json',
-          :host     => @queue_host,
-          :port     => @queue_port,
-          :protocol => :Kafka,
         }
       end
 
--- a/lib/topological_inventory/ansible_tower/messaging_client.rb
+++ b/lib/topological_inventory/ansible_tower/messaging_client.rb
@@ -3,7 +3,7 @@ require "manageiq-messaging"
 module TopologicalInventory
   module AnsibleTower
     class MessagingClient
-      OPERATIONS_QUEUE_NAME  = "platform.topological-inventory.operations-ansible-tower".freeze
+      OPERATIONS_QUEUE_NAME = "platform.topological-inventory.operations-ansible-tower".freeze
       REFRESH_QUEUE_NAME = "platform.topological-inventory.collector-ansible-tower".freeze
 
       # Kafka host name
--- a/lib/topological_inventory/ansible_tower/targeted_refresh/service_instance.rb
+++ b/lib/topological_inventory/ansible_tower/targeted_refresh/service_instance.rb
@@ -5,7 +5,7 @@ module TopologicalInventory
   module AnsibleTower
     module TargetedRefresh
       class ServiceInstance < TopologicalInventory::AnsibleTower::Cloud::Collector
-        REFS_PER_REQUEST_LIMIT = 20.freeze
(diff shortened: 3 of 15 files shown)


slemrmartin:
@syncrou I realized that targeted refresh is only new "operation". 
Question is, do we want to separate it as a new service(pod) (I made a new `bin/` script) or do we process it by operations worker?

EDIT: I would rather start it with new service and later join with full refresh service (see comment below)

slemrmartin:
This is a preview, but I want to know your opinions, @lindgrenj6, @syncrou 

Refresh can be ordered by new service (
https://github.com/slemrmartin/topological_inventory-scheduler/blob/refresh-by-one-job/lib/topological_inventory/scheduler/worker.rb)
(..later it can also schedule full and incremental refreshes so we can remove dependency on orchestrator and scale collectors in a standard way. )

The most important part of this PR is https://github.com/RedHatInsights/topological_inventory-ansible_tower/pull/99/files#diff-fe657fb90a0328ada5e347f8eb7e1f97. It's able to process one job by one query at this moment.

slemrmartin:
Update of ^^^:

Last commit allows update of multiple ServiceInstances (Jobs) with the same Source
Input is based on scheduler's master branch: https://github.com/slemrmartin/topological_inventory-scheduler/blob/master/lib/topological_inventory/scheduler/worker.rb

lindgrenj6:
So after finally giving this a look - I'd like to propose having the "scheduler" part be a k8s `CronJob` like the availability checks, mostly because it's just sitting there in a while loop sleeping and then doing one direct task: putting a message on a queue https://github.com/slemrmartin/topological_inventory-scheduler/blob/ce2ded46e6c57229c060a57ae7e6a7c619e72b5c/lib/topological_inventory/scheduler/worker.rb#L31

I think if we just had this as a cronjob that ran every minute it would be easier on resources (since we wouldn't have a pod just sitting there sleeping most of the time), and it would be easier to configure since we can just change the Cron schedule on the fly.

slemrmartin:
@lindgrenj6 I agree, I'm only afraid about the maintenance phase, you don't have `rsh` so you can't check the actual version of source code, the openshift UI seems to be much worse for jobs than pods and the openshift scheduler can be confused by some unknown problems (if you remember the problem with availability check cronjobs)

lindgrenj6:
@slemrmartin My only counters there are:

1) The sources availability checkers have been working fine for the last few months since we got the configuration ironed out with `startingDeadlineSeconds`, `concurrencyPolicy` etc, they're pretty solid.

2) If we really did need to rsh into a pod to troubleshoot stuff on production its relatively simple to just create a `Pod` resource from the `jobTemplate` section of the cronjob, from there we can make the entrypoint `tail -f /dev/null` then `rsh` in and poke around. 

Here is an example of a pod from my db backup/restore scripts for catalog/approval OCP prod: https://github.com/lindgrenj6/insights-rds-db-dump-container/blob/master/catalog/resources/catalog_dumper.yml 
We could run the same image SHA256 that the `CronJob` would be using in this case.

lib/topological_inventory/ansible_tower/collector/service_catalog.rb
```diff
@@ -1,3 +1,5 @@
+require 'pry-byebug'
```
lindgrenj6: Probably need to remove this.

lib/topological_inventory/ansible_tower/collector/service_catalog.rb
```diff
@@ -64,12 +66,17 @@ def get_service_offering_nodes(connection)
...
-      def get_service_instances(connection)
+      def get_service_instances(connection, refs: nil)
         fnc = lambda do |&block|
-          log_external_url("#{connection_manager.api_url(tower_hostname)}/#{connection.api.jobs.klass.endpoint}")
-          log_external_url("#{connection_manager.api_url(tower_hostname)}/#{connection.api.workflow_jobs.klass.endpoint}")
-          {:job          => connection.api.jobs.all(:page_size => limits[:service_instances]),
-           :workflow_job => connection.api.workflow_jobs.all(:page_size => limits[:service_instances])}.each_pair do |type, enumerator|
+          opts = {:page_size => limits[:service_instances]}
+          refs = refs.join(',') if refs.kind_of?(Array)
+          opts[:id__in] = refs if refs.present?
```
lindgrenj6: For this parameter I think it could be a little clearer if we had the assignment in a block rather than having multiple post-modifiers, e.g:

```ruby
if refs.kind_of?(Array)
  opts[:id__in] = refs.join(',')
end
```

lindgrenj6 (commented):
Just a few nits, I think I understand this well enough.

slemrmartin:
I think it's ready. Targeted refresh has now only connection to cloud towers, the on-premise can be implemented in next PR.

spec/targeted_refresh/processor_spec.rb
```diff
@@ -0,0 +1,96 @@
...
+  describe "#process" do
+    context "on implemented operation" do
+      let(:job_refs) { %w[10 20] }
+      let(:message) do
+        double(:message => 'ServiceInstance.refresh',
+               :payload => {
+                 :source_id  => '1',
+                 :source_uid => '3bfb8667-2b00-480b-bcbf-452bfb34a440',
+                 :params     => [{:task_id => '1', :source_ref => job_refs[0], :request_context => {"x-rh-identity": "xxx", "x-rh-insights-request-id": "111"}},
+                                 {:task_id => '2', :source_ref => job_refs[1], :request_context => {"x-rh-identity": "yyy", "x-rh-insights-request-id": "222"}}]
```
gmcculloug: Rubocop pointed out these two lines for "Use hash rockets" which I would recommend resolving.  These lines are using mixed syntax which we should avoid.

syncrou:
PR to add the new Kafka topic to Prod / Stage https://github.com/RedHatInsights/platform-mq/pull/80

lib/topological_inventory/ansible_tower/collector/service_catalog.rb
```diff
@@ -120,20 +121,28 @@ def get_tower_objects(connection, query_params, tower_types, on_premise: false,
...
           #
           # Logging Tower Full path
-          log_external_url("#{connection_manager.api_url(tower_hostname)}/#{api_obj.klass.endpoint}")
+          log_external_url("#{connection_manager.api_url(tower_hostname)}/#{slug || api_obj.klass.endpoint}")
           #
           # Calling Tower API
-          if on_premise
-            api_obj.all(query_params, receptor_params)
+          if slug
+            if on_premise
```
syncrou, line 135: @slemrmartin My understanding was `if slug` that was a guarantee we were `on_premise`.

lib/topological_inventory/ansible_tower/messaging_client.rb
```diff
@@ -0,0 +1,82 @@
+require "manageiq-messaging"
+
+module TopologicalInventory
+  module AnsibleTower
+    class MessagingClient
+      OPERATIONS_QUEUE_NAME = "platform.topological-inventory.operations-ansible-tower".freeze
```
syncrou, line 6: should these constants be environment variables?

lib/topological_inventory/ansible_tower/messaging_client.rb
```diff
@@ -0,0 +1,82 @@
...
+      def targeted_refresh_listener
+        @targeted_refresh_listener ||= ManageIQ::Messaging::Client.open(targeted_refresh_listener_opts)
+      end
+
+      def worker_listener_queue_opts
+        {
+          :auto_ack    => false,
+          :max_bytes   => 50_000,
+          :service     => OPERATIONS_QUEUE_NAME,
+          :persist_ref => "topological-inventory-operations-ansible-tower"
```
syncrou, line 45: Is this going to be the hostname like we're using in the other collectors?

lib/topological_inventory/ansible_tower/messaging_client.rb
```diff
@@ -0,0 +1,82 @@
...
+        }
+      end
+
+      def targeted_refresh_listener_queue_opts
+        {
+          :service     => REFRESH_QUEUE_NAME,
+          :persist_ref => "topological-inventory-collector-ansible-tower"
```
syncrou, line 52: Same about the hostname as above.

lib/topological_inventory/ansible_tower/targeted_refresh/service_instance.rb
```diff
@@ -0,0 +1,202 @@
...
+          # Get all tasks
+          tasks, refresh_state_uuid = {}, SecureRandom.uuid
+          params.to_a.each do |task|
+            if task['task_id'].blank? || task['source_ref'].blank?
+              logger.warn("ServiceInstance#refresh - missing data for task: #{task}")
+              next
+            end
+            tasks[task['task_id']] = task['source_ref']
+
+            if tasks.length == REFS_PER_REQUEST_LIMIT
```
syncrou, line 36: Is this tied to a paging limit of some type? Not sure why the number 20 is the default.

lib/topological_inventory/ansible_tower/targeted_refresh/service_instance.rb
```diff
@@ -0,0 +1,202 @@
...
+        attr_accessor :params, :source_id
+
+        def required_params
+          %w[source_id source params]
+        end
+
+        def params_missing?
+          is_missing = false
+          required_params.each do |attr|
+            if (is_missing = send(attr).blank?)
```
syncrou, line 68: Is the `if (is_missing = send(attr).blank?)` pattern used for readability?  I'm not sure if  `if(send(attr).blank?)` would be harder to read since we're iterating `required_params`.

lib/topological_inventory/ansible_tower/targeted_refresh/service_instance.rb
```diff
@@ -0,0 +1,202 @@
...
+          }
+
+          if on_premise?
+            refresh_part_on_premise(tasks_id, refresh_state_uuid, query_params)
+          else
+            refresh_part_cloud(tasks_id, refresh_state_uuid, refresh_state_part_uuid, query_params)
+          end
+        end
+
+        def refresh_part_on_premise(_tasks_id, refresh_state_uuid, query_params)
```
syncrou, line 93: is there a reason `refresh_part_on_premise` is not public vs `refresh_part_cloud` is?

lib/topological_inventory/ansible_tower/targeted_refresh/service_instance.rb
```diff
@@ -0,0 +1,202 @@
...
+        # Queries Sources API in the context of first task
+        def identity
+          @identity ||= params.to_a.first['request_context']
+        end
+
+        def api_client
+          @api_client ||= TopologicalInventory::Providers::Common::Operations::SourcesApiClient.new(identity)
+        end
+
+        # TODO: Join with PR #112 (ordering)
```
syncrou, line 174: Great idea!

lib/topological_inventory/ansible_tower/targeted_refresh/worker.rb
```diff
@@ -0,0 +1,67 @@
...
+          raise
+        rescue => err
+          tasks_id = if payload
+                       ids = payload['params'].to_a.collect { |task| task['task_id'] }
+                       ids.compact!
+                     end
+          logger.error("#{model}##{method} - Task[ id: #{tasks_id.to_a.join(' | id: ')} ] #{err.cause}\n#{err}\n#{err.backtrace.join("\n")}")
+          raise
+        ensure
+          message.ack
```
syncrou, line 57: Is this going to have to change to default `ack` so it is inline with the receptor client issues we were seeing?

lib/topological_inventory/ansible_tower/targeted_refresh/service_instance.rb
```diff
@@ -0,0 +1,202 @@
...
+                                                                                     refresh_state_uuid,
+                                                                                     refresh_state_started_at,
+                                                                                     :sweeping_enabled => false)
+          get_service_instances(connection, query_params,
+                                :on_premise        => true,
+                                :receptor_receiver => receiver,
+                                :receptor_params   => receptor_params)
+        end
+
+        public
```
lindgrenj6: I think this method should be moved up (removing the public specifier) so we can just stick to one `private` definition rather than switching back and forth.

lib/topological_inventory/ansible_tower/targeted_refresh/service_instance.rb
```diff
@@ -0,0 +1,202 @@
...
+                          end
+        end
+
+        def on_premise?
+          @on_premise ||= endpoint.receptor_node.to_s.strip.present?
+        end
+
+        def set_connection_data!
+          raise TopologicalInventory::Providers::Common::Operations::Source::ERROR_MESSAGES[:endpoint_not_found] unless endpoint
+          raise TopologicalInventory::Providers::Common::Operations::Source::ERROR_MESSAGES[:authentication_not_found] unless authentication
```
lindgrenj6: An authentication won't be present on receptor-enabled towers, at least not currently. So it shouldn't raise this unless we're not `on_premise?`

lib/topological_inventory/ansible_tower/targeted_refresh/service_instance.rb
```diff
@@ -0,0 +1,202 @@
...
+
+            if tasks.length == REFS_PER_REQUEST_LIMIT
+              refresh_part(tasks, refresh_state_uuid, SecureRandom.uuid)
+              tasks = {}
+            end
+          end
+
+          refresh_part(tasks, refresh_state_uuid, SecureRandom.uuid) unless tasks.empty?
+        rescue => err
+          logger.error("ServiceInstance#refresh - Error: #{err.message}\n#{err.backtrace.join("\n")}")
```
lindgrenj6, line 44: Can we get some metrics around this? Would be really nice to alert when `ServiceInstance#refresh` is failing a lot.

lib/topological_inventory/ansible_tower/targeted_refresh/service_instance.rb
```diff
@@ -0,0 +1,202 @@
...
+        end
+
+        def set_connection_data!
+          raise TopologicalInventory::Providers::Common::Operations::Source::ERROR_MESSAGES[:endpoint_not_found] unless endpoint
+          raise TopologicalInventory::Providers::Common::Operations::Source::ERROR_MESSAGES[:authentication_not_found] unless authentication
+
+          self.tower_hostname = full_hostname(endpoint)
+        end
+
+        # TODO: Join with other operations
```
lindgrenj6, line 150: Yeah I think we could make use of the `Source` class in providers-common here, if we made the endpoint/auth methods public, would be nice to store the state there, definitely a follow-up though when we refactor a lot of the availability checking stuff like in #112.

lindgrenj6 (commented):
Getting close on this one!

lib/topological_inventory/ansible_tower/targeted_refresh/service_instance.rb
slemrmartin: @lindgrenj6 targeted refresh is for both cloud and on-premise

lib/topological_inventory/ansible_tower/targeted_refresh/service_instance.rb
slemrmartin: I guess I added some TODO, I think this PR is big enough, so I'll rather add metrics in next PR

lib/topological_inventory/ansible_tower/collector/service_catalog.rb
slemrmartin: @syncrou not always, if you want to use "find_all_by_url" (it's also ansible-tower-client's method), you need to provide slug

lib/topological_inventory/ansible_tower/messaging_client.rb
slemrmartin: this is a refactoring to have all topic's data together. I don't think we want to expose everything as an env variable. 
Moreover auto-creating topic is disabled, it doesn't make sense to me to have it configurable

lib/topological_inventory/ansible_tower/messaging_client.rb
slemrmartin: this is not an opts for receptor! This is constant consumer group shared accross ansible-tower-operations replicas

lib/topological_inventory/ansible_tower/messaging_client.rb
slemrmartin: the same answer as above :) that's not receptor's listener

lib/topological_inventory/ansible_tower/targeted_refresh/service_instance.rb
slemrmartin: @syncrou I just specified something. The only limitation is max length of URL (which can't be computed in on-premise mode).
I guess 100 could be also fine. What do you suggest?

lib/topological_inventory/ansible_tower/targeted_refresh/service_instance.rb
slemrmartin: both of them are private. Hm, obviously it's not a good idea to include public block inside private methods, I'll move it

lib/topological_inventory/ansible_tower/targeted_refresh/service_instance.rb
slemrmartin: I wanted to group related methods together, but it seems it's not a good idea

lib/topological_inventory/ansible_tower/targeted_refresh/worker.rb
slemrmartin: This is aligned with operations listener style. As the operation can take longer time, you can't have auto-ack because it can auto-timeout then. 
The receptor client issues were related to listener without `persist_ref`

lib/topological_inventory/ansible_tower/targeted_refresh/service_instance.rb
slemrmartin: @syncrou this method is duplicated accross operations, so it'll be moved to providers-common in next refactoring. 
But I'm not sure what you are suggesting

lib/topological_inventory/ansible_tower/targeted_refresh/service_instance.rb
slemrmartin: it'll wait to next PR because there are more duplicates accross operations

lib/topological_inventory/ansible_tower/targeted_refresh/service_instance.rb
lindgrenj6: Yes, we just need to be checking in the `fetch_authentication` method if we're on premise and return a string or something (I just return a string constant in common) that way it doesn't raise the exception for receptor towers.

lib/topological_inventory/ansible_tower/targeted_refresh/service_instance.rb
syncrou: 20 is probably fine. I was more interested in now that came to be. Seems like it can't be computed so any number as a default is okay.

lib/topological_inventory/ansible_tower/targeted_refresh/service_instance.rb
syncrou: I don't see the need for the extra boolean:
```
          required_params.each do |attr|
            if (send(attr).blank?)
```

Probably worth a discussion later. Nothing to hold up this PR.

syncrou (commented):
This looks good so far. A few items we can update in followup PRS.

There are a couple concerns @lindgrenj6 had yet, so I will wait till those are fully cleared up to approve.

lib/topological_inventory/ansible_tower/targeted_refresh/service_instance.rb
slemrmartin: fixed

slemrmartin:
@syncrou there is one possible issue: If this PR removes collecting of all service_instances, then ordering won't be completed until the cron job in e2e_deploy will be merged too. 

Especially I'm not sure how it is synchronized between standard code and e2e_deploy templates => app_interface

EDIT: Regarding that, won't it be more safe to remove the full collecting by following PR?

lindgrenj6 (approved):

PR #100, opened 2020-06-23
Respect Endpoint port setting on Operations

lindgrenj6:
https://bugzilla.redhat.com/show_bug.cgi?id=1850116

cc @gmcculloug @syncrou @slemrmartin 

~~Not the direct cause of our issue - but did find this while debugging. The collector respects whatever port is configured in pool-mode but NOT in single-collector mode.~~

1. So after doing some digging, I found that the culprit is that on the Ansible Tower Operations Source class was failing to set up the host correctly from the sources endpoint response, it was ignoring the port. 

2. I also found that there was no CLI option for port - so I added that in a tiny commit as well.

### DIFF
--- a/bin/ansible_tower-collector
+++ b/bin/ansible_tower-collector
@@ -22,6 +22,7 @@ require require_dev_path if File.exist?(require_dev_path)
 # --source      (ENV["SOURCE_UID"])      - core: Source.uid (manager for this collector)
 # --scheme      (ENV["ENDPOINT_SCHEME"]) - ansible tower scheme, https by default
 # --host        (ENV["ENDPOINT_HOST"])   - ansible tower url
+# --port        (ENV["ENDPOINT_PORT"])   - ansible tower port
 # --user        (ENV["AUTH_USERNAME"])   - credentials: username
 # --password    (ENV["AUTH_PASSWORD"])   - credentials: password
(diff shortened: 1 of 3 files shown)


lindgrenj6:
@gmcculloug @syncrou this is ready for review.

syncrou (approved):
This is great! Thanks for the quick turn around. LGTM 👍

gmcculloug:
Not knowing this repo it looks like the current specs are only testing the change to `lib/topological_inventory/ansible_tower/operations/source.rb` and not the `bin/ansible_tower-collector`.  Otherwise, the changes look good to me.

lindgrenj6:
@gmcculloug that's correct, the collector side of things is happening here: https://github.com/RedHatInsights/topological_inventory-ansible_tower/blob/93d0793e2f1b965104f2b613178ed03ba185d209/lib/topological_inventory/ansible_tower/collectors_pool.rb#L32-L34 

the collector side of things takes the config file's port setting and feeds it to the ansible tower client there, the main difference is that we were going off of a source authentication response rather than a config file, so it was missed.

Personally I blame the Tower Client API for taking `host` and no port - it could probably be refactored eventually: https://github.com/RedHatInsights/topological_inventory-ansible_tower/blob/93d0793e2f1b965104f2b613178ed03ba185d209/lib/topological_inventory/ansible_tower/connection.rb#L8

PR #73, opened 2020-03-05
[WIP] On premise ordering

slemrmartin:
**Issue**: https://github.com/RedHatInsights/receptor_controller-client-ruby/issues/2

`ServiceOffering#order` operation for on-premise Towers (using receptor-client)

---

- [ ] **depends on**: https://github.com/RedHatInsights/topological_inventory-ansible_tower/pull/71 (needs rebase)
- [ ] **depends on**: https://github.com/RedHatInsights/receptor_controller-client-ruby/pull/5

PR #101, opened 2020-06-30
[WIP] on premise availability check

lindgrenj6:
# This is built off of Martin's #73 PR, My only change was the last commit. 

### All of this will most likely change a bit since it's based on a bunch of other work that we will be changing to use the catalog_receptor plugin, but at the end of the day at least the availability checking won't be a huge change. 

Common changes in: https://github.com/RedHatInsights/topological_inventory-providers-common/pull/36

cc @syncrou @slemrmartin 

----

- Receptor API client
- Receptor response worker as a class variable
- Launch Job Template
- Default response timeout
- Receptor client: from concept to real implementation
- Receptor service_credentials collecting
- Collector using ReceptorClient
- Support for ServiceCredentials join tables
- Common identifier to receptor log messages
- Order operation through receptor
- Availability check w/ receptor node


lindgrenj6:
Closing in favor of #107

PR #102, opened 2020-07-03
Support for on-premise towers

slemrmartin:
**Issue** https://github.com/RedHatInsights/receptor_controller-client-ruby/issues/2

Support for communication with AnsibleTower through Cloud Receptor Controller.

Included API Client provides responses in [AnsibleTowerClient](https://github.com/ansible/ansible_tower_client_ruby)'s compatible format. Supports both synchronous and asynchronous calls. 

**Doesn't affect current functionality** (modified collector is implemented by following PR #71)

---

- [x] **depends on** https://github.com/RedHatInsights/receptor_controller-client-ruby/pull/5
- [x] **depends on** https://github.com/RedHatInsights/receptor_controller-client-ruby/pull/13


bin/ansible_tower-collector
```diff
@@ -73,16 +69,30 @@ ingress_api_uri = URI(args[:ingress_api])
...
+  config.controller_host   = "#{args[:receptor_controller_host]}:#{args[:receptor_controller_port]}"
+  config.queue_host        = args[:queue_host]
+  config.queue_port        = args[:queue_port]
+end
+
+
+metrics = TopologicalInventory::AnsibleTower::Collector::ApplicationMetrics.new(args[:metrics_port])
+if args[:config].nil?
+  collector = TopologicalInventory::AnsibleTower::Collector.new(args[:source],
+                                                                "#{args[:scheme]}://#{args[:host]}:#{args[:port]}",
```
lindgrenj6, line 85: We are going to need a config for receptor enabled collectors too, we won't have a URL/user/pass for on-prem sources.

lib/topological_inventory/ansible_tower/receptor/api_object.rb
```diff
@@ -0,0 +1,210 @@
...
+
+        directive_type = async? ? :non_blocking : :blocking
+        directive = receptor_client.directive(connection.account_number,
+                                              connection.receptor_node,
+                                              :directive          => RECEPTOR_DIRECTIVE,
+                                              :log_message_common => payload['url'],
+                                              :payload            => payload.to_json,
+                                              :type               => directive_type)
+
+        if async?
```
lindgrenj6, line 105: Just a comment on future implementation:

--- 

To make things easier in the long run, I think we should only use the `:non_blocking` directive. 

I don't see why we would use `:blocking` mainly because we should not be storing state in the thread, so we should probably just do all-async for now until we do the refactor where we store the directive information in a redis cache.

lib/topological_inventory/ansible_tower/receptor/connection.rb
```diff
@@ -0,0 +1,62 @@
...
+        @receptor_endpoint = receptor_client.config.job_url
+      end
+
+      def default_api_path
+        "/api/v2".freeze
+      end
+
+      # org_id with any number is required by receptor_client controller
+      # TODO: Replace x-rh-identity with x-rh-rbac-psk
+      def identity_header(account = account_number)
```
lindgrenj6, line 41: Like I mentioned earlier, we can check for a PSK in the ENV and set the headers based on that and fall back if necessary to the x-rh-identity header.

https://github.com/dehort/platform-receptor-controller-1/blob/b5c3da14af004b0d1d21134a6a00a560147009ef/internal/controller/api/api.spec.json#L32-L34

lib/topological_inventory/ansible_tower/receptor/template.rb
```diff
@@ -0,0 +1,27 @@
...
+
+      # URI + String removes URI path if String starts with '/'
+      # This technique is used by Faraday client
+      #
+      # Response from https://18.188.178.206/api/v2/api/v2/job_templates/9/survey_spec/ failed: HTTP status: 404"}
+      def survey_spec
+        spec_url = related['survey_spec']
+        return nil unless spec_url
+        api.get(spec_url).body
+      rescue AnsibleTowerClient::UnlicensedFeatureError
```
lindgrenj6: Is this error only here when they try to use the feature on AWX or something? Either way I think we should log an error that way if a customer notices it isn't working we'll at least have some idea why.

lib/topological_inventory/ansible_tower/collector.rb
```diff
@@ -55,7 +55,7 @@ def service_catalog_entity_types
 
     # Connection to endpoint (for each entity type the same)
     def connection_for_entity_type(_entity_type)
-      connection_manager.connect(tower_hostname, tower_user, tower_passwd)
+      connection_manager.connect(:base_url => tower_hostname, :username => tower_user, :password => tower_passwd)
```
lindgrenj6, line 58: Once again probably going to need a case here for receptor-enabled towers.

lindgrenj6 (commented):
This is a good first-pass that will work with the current receptor-client-controller implementation. Just a few change requests, but this is also WIP so I may be jumping the gun.

bin/ansible_tower-collector
slemrmartin: yes, I splitted the functionality to two PRs, this PR doesn't affect collecting only adds support for on-premise.
What you mentioned is implemented in https://github.com/RedHatInsights/topological_inventory-ansible_tower/pull/71/files#diff-25c276be90feb3bc2d1b625302a95c0aR89

lib/topological_inventory/ansible_tower/receptor/api_object.rb
slemrmartin: yes, as I posted in the receptor-client PR, blocking directives are now used for subqueries (config/survey/node credentials) as it needs more effort to convert them to async. So it should be implemented in following PRs

lib/topological_inventory/ansible_tower/receptor/template.rb
slemrmartin: Frankly I just copied this rescue from AnsibleTowerClient repo, I haven't seen it before. I'll add logging

lib/topological_inventory/ansible_tower/collector.rb
slemrmartin: solved in following PR by inherited `Receptor::Collector` class

lib/topological_inventory/ansible_tower/receptor/template.rb
lindgrenj6: All good! haha I don't know why it wasn't handled before either.

lib/topological_inventory/ansible_tower/connection_manager.rb
```diff
@@ -0,0 +1,73 @@
...
+  # depending on provided parameters
+  class ConnectionManager
+    include Logging
+
+    delegate :api_url, :to => :connection
+
+    attr_reader :connection
+
+    @@sync = Mutex.new
+    @@receptor_client = nil
```
roliveri: I think it would be better to make these class instance variables as opposed to class variables.
This will protect them from being inadvertently overridden from instances and subclasses.

lib/topological_inventory/ansible_tower/receptor/connection.rb
slemrmartin: @lindgrenj6 Is it possible that psk will not work and identity header will? Is the PSK applicable to development?

lindgrenj6: Good question, I think we should probably use the PSK if the ENV var is set, but fall back to x-rh-identity that way we can still use it in development mode locally. Does that sound ok?

slemrmartin:
@syncrou @lindgrenj6 unwipped, ready. 
I would like to have merged this before #99 as there will be conflicts

lindgrenj6 (approved):
LGTM :steam_locomotive: lets get this receptor chugging along!

roliveri (approved):

syncrou (approved):
LGTM 👍

PR #104, opened 2020-07-15
update cli queue_port option type to integer

lindgrenj6:
Pods in CI are in crashloopbackoff I assume due to the dc's not having the queue_port value since they haven't been re-created. This will fix that.


syncrou (approved):
LGTM 👍 Thanks for the quick turn around on this on @lindgrenj6!

PR #105, opened 2020-07-16
Fixed WorkflowJobTemplate parsing

slemrmartin:
This is really tricky bug. `WorkflowJobTemplate` doesn't have all the extra attributes like `JobTemplate`. 

But in `AnsibleTowerClient`, the WorkflowJobTemplate is inherited from JobTemplate. 
JobTemplate defines all these methods dynamically at class level, so they are then accessible by WorkflowJobTemplate. 

We haven't discovered this in public tower collecting because collecting of JobTemplates is always preceding to collecting of WorkflowJobTemplates, so methods are defined for both classes. 

In case of asynchronous collecting, the WorkflowJobTemplate can be parsed first and then raises error. 

---

[TPINVTRY-1012](https://projects.engineering.redhat.com/browse/TPINVTRY-1012)

### DIFF
--- a/lib/topological_inventory/ansible_tower/parser/service_offering.rb
+++ b/lib/topological_inventory/ansible_tower/parser/service_offering.rb
@@ -6,25 +6,32 @@ module TopologicalInventory::AnsibleTower
 
         service_inventory = lazy_find(:service_inventories, :source_ref => template.inventory_id.to_s) if template.respond_to?(:inventory_id)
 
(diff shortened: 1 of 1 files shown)


lindgrenj6 (approved):
This looks like it was a fun one to track down.

syncrou (approved):
Nice find @slemrmartin! Happy we found this now.

LGTM 👍

PR #107, opened 2020-07-24
Add on-premise availability check through receptor

lindgrenj6:
**DEPENDS ON:**
- [x] https://github.com/RedHatInsights/topological_inventory-providers-common/pull/36
- [x] [1.0.7 Providers Common Released](https://github.com/RedHatInsights/topological_inventory-providers-common/pull/42)
- [x] https://github.com/RedHatInsights/e2e-deploy/pull/2026

cc @syncrou @slemrmartin 

\# TODO: 
- [x] specs

### DIFF
--- a/lib/topological_inventory/ansible_tower/operations/source.rb
+++ b/lib/topological_inventory/ansible_tower/operations/source.rb
@@ -11,8 +11,13 @@ module TopologicalInventory
         private
 
         def connection_check
-          connection = ::TopologicalInventory::AnsibleTower::Cloud::Connection.new
-          connection = connection.connect(full_hostname(endpoint), authentication.username, authentication.password)
+          connection = ::TopologicalInventory::AnsibleTower::ConnectionManager.new(source_id).connect(
+            :base_url       => full_hostname(endpoint),
+            :username       => authentication.try(:username),
+            :password       => authentication.try(:password),
+            :receptor_node  => endpoint.receptor_node,
+            :account_number => account_number
+          )
           connection.api.version
 
           [STATUS_AVAILABLE, nil]
--- a/lib/topological_inventory/ansible_tower/operations/worker.rb
+++ b/lib/topological_inventory/ansible_tower/operations/worker.rb
@@ -2,6 +2,7 @@ require "manageiq-messaging"
 require "topological_inventory/ansible_tower/logging"
 require "topological_inventory/ansible_tower/operations/processor"
 require "topological_inventory/ansible_tower/operations/source"
+require "topological_inventory/ansible_tower/connection_manager"
 
 module TopologicalInventory
   module AnsibleTower
@@ -29,6 +30,7 @@ module TopologicalInventory
           logger.error("#{err.cause}\n#{err.backtrace.join("\n")}")
         ensure
           client&.close
+          TopologicalInventory::AnsibleTower::ConnectionManager.stop_receptor_client
         end
 
         private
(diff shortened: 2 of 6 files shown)


gmcculloug:
cc @mkanoor

lindgrenj6:
@syncrou @slemrmartin this is ready for review, as well as the dependencies

lib/topological_inventory/ansible_tower/operations/worker.rb
```diff
@@ -14,6 +16,11 @@ def initialize(messaging_client_opts = {})
         end
 
         def run
+          # Start up the receptor worker
+          receptor_client = ReceptorController::Client.new(:logger => logger)
+          TopologicalInventory::AnsibleTower::ConnectionManager.class_variable_set(:@receptor_client, receptor_client)
+          receptor_client.start
```
slemrmartin: I think we don't need new receptor_client for each request, it was intended as one client per pod (it means one kafka listener per process). The `logger` will be the same in both cases.
Then you don't need to propagate it to all the methods.

lib/topological_inventory/ansible_tower/operations/source.rb
```diff
@@ -11,9 +11,19 @@ class Source < TopologicalInventory::Providers::Common::Operations::Source
...
+              endpoint.receptor_node,
+              account_number
+            )
+
+            connection.api.config
+          else
+            connection = ::TopologicalInventory::AnsibleTower::Cloud::Connection.new
+            connection = connection.connect(full_hostname(endpoint), authentication.username, authentication.password)
+            connection.api.version
+          end
```
slemrmartin: `ConnectionManager.connect` is doing this, with lazy initialization of `receptor_client` (except the api call)

Why did you choose `api.config` instead of `api.version` for on-premise?

lindgrenj6: Cool! I updated it to use the `ConnectionManager.connect`, and I did api.version for both. I just did config out of habit. They both use the same API call now.

lib/topological_inventory/ansible_tower/operations/worker.rb
lindgrenj6: :+1: pulled the receptor out of everything, the state is all handled in the ConnectionManager class.

lib/topological_inventory/ansible_tower/operations/source.rb
```diff
@@ -11,8 +11,13 @@ class Source < TopologicalInventory::Providers::Common::Operations::Source
         private
 
         def connection_check
-          connection = ::TopologicalInventory::AnsibleTower::Cloud::Connection.new
-          connection = connection.connect(full_hostname(endpoint), authentication.username, authentication.password)
+          connection = ::TopologicalInventory::AnsibleTower::ConnectionManager.new(source_id).connect(
```
slemrmartin, line 14: I'm not sure if it'll work in all cases. Either make if/else there, or add `.present?` to checks there: https://github.com/RedHatInsights/topological_inventory-ansible_tower/blob/f42cda8cbd05b9360622ba683c5814177a4d5dcb/lib/topological_inventory/ansible_tower/connection_manager.rb#L47

slemrmartin (approved):
:+1: nice work!

syncrou (approved):
LGTM 👍 

Only partial comment was around the need for `.present?` on `receptor_node`. After discussing with @lindgrenj6 / @slemrmartin that there could be a `""` value passed vs `nil` it makes sense to me.

PR #110, opened 2020-07-29
Adding Directive double to the source_spec

slemrmartin:
Following #107 

Adding response to "receptor_client.directive" as a double not to raise exception.

The spec has been working because rescue block is catching the exception and this spec is focusing on something else.


slemrmartin:
cc @mkanoor

lindgrenj6 (approved):
LGTM, it's funny the spec was succeeding because I was `expect`ing the call to be made, and it did, but it also threw an exception :grimacing:

syncrou (approved):
LGTM 👍

PR #114, opened 2020-08-04
Starting receptor worker immediately

slemrmartin:
When reading messages from operations worker listener, receptor kafka listener should be initialized. This PR moves start from lazy initialization to the beginning of the process.

Also added the same for receptor controller

---

**Depends on**
- [x] https://github.com/RedHatInsights/receptor_controller-client-ruby/pull/20 
- [x] https://github.com/RedHatInsights/receptor_controller-client-ruby/pull/28


lindgrenj6 (approved):

syncrou (approved):
LGTM 👍

PR #115, opened 2020-08-05
Fix sending receptor response array

slemrmartin:
The Receptor::AsyncReceiver received only first record of tower response


lindgrenj6 (approved):
Tested locally with multiple towers, seems to be working great!

syncrou (approved):
LGTM 👍

PR #116, opened 2020-08-05
Collector: waiting for async responses in main thread

slemrmartin:
When using async responses from receptor, request threads are done very quickly. 
So the main thread should wait until all responses are received

### DIFF
--- a/lib/topological_inventory/ansible_tower/receptor/collector.rb
+++ b/lib/topological_inventory/ansible_tower/receptor/collector.rb
@@ -7,6 +7,8 @@ module TopologicalInventory::AnsibleTower
       def initialize(source, receptor_node, account_number, metrics, standalone_mode: true)
         super(source, metrics, :standalone_mode => standalone_mode)
         self.account_number = account_number # eq Tenant.external_tenant or account_number in x-rh-identity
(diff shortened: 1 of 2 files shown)


lib/topological_inventory/ansible_tower/receptor/collector.rb
```diff
@@ -35,20 +37,48 @@ def async_collecting_finished(entity_type, refresh_state_uuid, total_parts)
...
+
+        super
+      end
+
+      def wait_for_collected_data
+        super
+
+        # Wait until all async responses are sent
+        while entity_types_collected_cnt.value < service_catalog_entity_types.size
+          if last_save_at.nil? || (last_save_at < 2.minutes.ago && last_save_at >= 10.minutes.ago)
```
lindgrenj6: I think we should make this value `10.minutes.ago` configurable, that way we can tweak it as time goes on, some towers might not be able to be collected fully within 10 minutes.

slemrmartin: I agree, but this value means that there was at least one successful data response in last 10 minutes. Then it resets the 10 minutes counter

lindgrenj6 (approved):
LGTM, @syncrou can we get this merged?

syncrou (approved):
LGTM 👍

PR #118, opened 2020-08-11
Removed collecting of service_instance_nodes

slemrmartin:
We are collecting ServiceInstanceNodes (Workflow Job Nodes). 
It requires most requests but afaik we don't need this data

So I think we don't need to wait for targeted refresh for removing them.


slemrmartin:
cc @mkanoor

syncrou (approved):
LGTM 👍

PR #119, opened 2020-08-13
Fixed receptor collector waiting when everything timeouts

slemrmartin:
Fix of #116. 
When all responses timeout or return error, it ended in a infinite loop

### DIFF
--- a/lib/topological_inventory/ansible_tower/receptor/collector.rb
+++ b/lib/topological_inventory/ansible_tower/receptor/collector.rb
@@ -8,7 +8,7 @@ module TopologicalInventory::AnsibleTower
         super(source, metrics, :standalone_mode => standalone_mode)
         self.account_number = account_number # eq Tenant.external_tenant or account_number in x-rh-identity
         self.entity_types_collected_cnt = Concurrent::AtomicFixnum.new(0)
-        self.last_save_at = nil
+        self.last_response_at = nil
         self.max_wait_sync_threshold = ENV['RECEPTOR_COLLECTOR_MAX_WAIT_SYNC'] || 10
         self.receptor_node  = receptor_node
         self.tower_hostname = "receptor://#{receptor_node}" # For logging
@@ -38,7 +38,6 @@ module TopologicalInventory::AnsibleTower
       end
 
       def async_save_inventory(refresh_state_uuid, parser)
-        self.last_save_at = Time.now.utc
         refresh_state_part_collected_at = Time.now.utc
         refresh_state_part_uuid = SecureRandom.uuid
         save_inventory(parser.collections.values, inventory_name, schema_name, refresh_state_uuid, refresh_state_part_uuid, refresh_state_part_collected_at)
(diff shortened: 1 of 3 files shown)


lib/topological_inventory/ansible_tower/receptor/collector.rb
```diff
@@ -53,12 +52,18 @@ def async_sweep_inventory(refresh_state_uuid, sweep_scope, total_parts, refresh_
         logger.sweeping(:finish, source, sweep_scope, refresh_state_uuid)
       end
 
+      def response_received!
+        self.last_response_at = Time.now.utc
```
slemrmartin, line 56: Although it's not thread safe assignment, in the worst case the value can be different by miliseconds. So it doesn't matter

lib/topological_inventory/ansible_tower/receptor/collector.rb
```diff
@@ -53,12 +52,18 @@ def async_sweep_inventory(refresh_state_uuid, sweep_scope, total_parts, refresh_
         logger.sweeping(:finish, source, sweep_scope, refresh_state_uuid)
       end
 
+      def response_received!
```
gmcculloug, line 55: What is the "!" signifying in the method name?

I know in many Rails methods this is used to say that the method will raise an error if something fails, but it is not clear to me why it is used here.

slemrmartin: "!" is also used in ruby methods to pointing out that the method is changing internal state of object

lib/topological_inventory/ansible_tower/receptor/collector.rb
```diff
@@ -69,12 +74,18 @@ def wait_for_collected_data
 
         # Wait until all async responses are sent
         while entity_types_collected_cnt.value < service_catalog_entity_types.size
-          if last_save_at.nil? || (last_save_at < ReceptorController::Client::Configuration.default.response_timeout.minutes.ago && last_save_at >= max_wait_sync_threshold.minutes.ago)
+          if last_response_at.nil? && refresh_started_at >= max_wait_sync_threshold.minutes.ago
+            logger.warn("[ASYNC] Collector for source_uid: #{source}: No response received since #{refresh_started_at}")
+            sleep(60)
+          elsif last_response_at < ReceptorController::Client::Configuration.default.response_timeout.minutes.ago && last_response_at >= max_wait_sync_threshold.minutes.ago
```
gmcculloug: Will this condition throw an error if `last_response_at` is `nil`?  (Same for the last `elsif` block)

```ruby
nil < 100
=> NoMethodError (undefined method `<' for nil:NilClass)
```

Should there be a higher level condition here that checks for `if last_response_at` and then do further processing inside of that?

```ruby
while entity_types_collected_cnt.value < service_catalog_entity_types.size
  if last_response_at
    ...
  else
    ...
  end
end
```

slemrmartin: ou, you're right

roliveri (approved):

syncrou (approved):
LGTM 👍

PR #122, opened 2020-08-19
Fixed Ordering Issue

mkanoor:
1. There was an extra key called **data** that was being inserted during
ordering which was being rejected by the Ansible Tower.
2. When sending parameters to the receptor plugin you don't have to
put the query parameters in the href_slug you can send it separately.
The appending of the query parameters to the slug has been removed its
redundant.

### DIFF
--- a/lib/topological_inventory/ansible_tower/receptor/api_object.rb
+++ b/lib/topological_inventory/ansible_tower/receptor/api_object.rb
@@ -29,7 +29,7 @@ module TopologicalInventory::AnsibleTower
       end
 
       def post(data)
-        response = send_request(:post, endpoint, :data => data)
+        response = send_request(:post, endpoint, data)
         raw_kafka_response(response)
       end
 
@@ -82,7 +82,6 @@ module TopologicalInventory::AnsibleTower
 
       def build_payload(http_method, path, params = nil, receptor_opts = {})
         slug = path.to_s
-        slug += "?#{params.to_query}" if params
         payload = {
           'method'          => http_method.to_s.upcase,
           'href_slug'       => slug,
--- a/spec/receptor/api_object_spec.rb
+++ b/spec/receptor/api_object_spec.rb
@@ -105,7 +105,7 @@ RSpec.describe TopologicalInventory::AnsibleTower::Receptor::ApiObject do
                                      .and_return(api_object))
 
         expect(api_object).to(receive(:send_request)
-                                .with(:post, job_order_path, :data => post_data)
+                                .with(:post, job_order_path, post_data)
                                 .and_return('status' => 200, 'body' => body))
 
         response = api.post(job_order_path, post_data)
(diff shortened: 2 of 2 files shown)


mkanoor:
@slemrmartin @syncrou @lindgrenj6 
Please review

lib/topological_inventory/ansible_tower/receptor/api_object.rb
```diff
@@ -29,7 +29,7 @@ def get
       end
 
       def post(data)
-        response = send_request(:post, endpoint, :data => data)
+        response = send_request(:post, endpoint, data)
```
syncrou, line 32: @lindgrenj6 @slemrmartin was line 32 `:data => data` in the cloud tower request? I'm interested why it was here if Tower didn't like the request. Unless the difference here is how the href_slug is handled vs the direct url.

mkanoor: @syncrou 
https://github.com/ansible/ansible_tower_client_ruby/blob/47155754e93da0d00afaeaac1a54681270a52663/lib/ansible_tower_client/base_models/job_template.rb#L16

lib/topological_inventory/ansible_tower/receptor/api_object.rb
```diff
@@ -82,7 +82,6 @@ def endpoint
 
       def build_payload(http_method, path, params = nil, receptor_opts = {})
         slug = path.to_s
-        slug += "?#{params.to_query}" if params
```
slemrmartin, line 85: @mkanoor why? How can you send the "id__in=" and other filters?

spec/receptor/api_object_spec.rb
```diff
@@ -192,7 +192,7 @@
 
         payload = {
           'method'          => 'GET',
-          'href_slug'       => '/api/v2/job_templates?id__in=42%2C1',
```
slemrmartin, line 195: why?

lib/topological_inventory/ansible_tower/receptor/api_object.rb
slemrmartin: or page_size?

lib/topological_inventory/ansible_tower/receptor/api_object.rb
slemrmartin: ah, the main difference is, that it was intended that `AnsibleTower::Receptor::Template` will launch, but `AnsibleTowerClient::JobTemplate` is used instead (as a result of previous `find` method). 

Anyway the `params` are intended as string after `?` in href_slug (GET requests) and `:data` should be sent in the payload as "params" (it is intended for POST data, @mkanoor, right?)

slemrmartin:
I prepared #123, it comes with the separation of query params and POST data as it was the original intention

lib/topological_inventory/ansible_tower/receptor/api_object.rb
mkanoor: @slemrmartin They all go into params. All the Query Values go into params. By adding it on the url line you are adding extra work which is not needed.

lib/topological_inventory/ansible_tower/receptor/api_object.rb
mkanoor: @slemrmartin We dont need anything with the href_slug. You can put all the parameters into params.

mkanoor:
Closed in favor of #123

PR #126, opened 2020-09-01
Add Kafka Heartbeat Thread

lindgrenj6:
Adding a heartbeat thread that runs alongside our process_message thread. This should hopefully keep kafka from kicking our consumer out. 

cc @syncrou @slemrmartin

JIRA: https://projects.engineering.redhat.com/browse/TPINVTRY-1060

### DIFF
--- a/lib/topological_inventory/ansible_tower/operations/worker.rb
+++ b/lib/topological_inventory/ansible_tower/operations/worker.rb
@@ -19,6 +19,7 @@ module TopologicalInventory
           TopologicalInventory::AnsibleTower::ConnectionManager.start_receptor_client
           # Open a connection to the messaging service
           client = ManageIQ::Messaging::Client.open(messaging_client_opts)
+          heartbeat_thread(client)
 
           logger.info("Topological Inventory AnsibleTower Operations worker started...")
           client.subscribe_topic(queue_opts) do |message|
@@ -31,6 +32,7 @@ module TopologicalInventory
         rescue => err
           logger.error("#{err.cause}\n#{err.backtrace.join("\n")}")
         ensure
+          heartbeat_thread.exit
           client&.close
           TopologicalInventory::AnsibleTower::ConnectionManager.stop_receptor_client
         end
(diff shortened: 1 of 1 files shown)


lib/topological_inventory/ansible_tower/operations/worker.rb
```diff
@@ -71,6 +73,16 @@ def default_messaging_opts
             :group_ref  => "topological-inventory-operations-ansible-tower"
           }
         end
+
+        # TODO: Probably move this to common eventually, if we need it elsewhere.
+        def heartbeat_thread(client = nil)
+          @heartbeat_thread ||= Thread.new do
+            loop do
+              sleep(15)
+              client.send(:topic_consumer, queue_opts[:persist_ref], nil).trigger_heartbeat!
```
slemrmartin, line 82: it should catch and recover from potential exception

slemrmartin (commented):
looks good, just a small thing

lib/topological_inventory/ansible_tower/operations/worker.rb
lindgrenj6: Updated!

lib/topological_inventory/ansible_tower/operations/worker.rb
```diff
@@ -71,6 +73,18 @@ def default_messaging_opts
...
+            rescue => e
```
syncrou: Is there a real issue with this placement? Rubocop seems to think so.

lindgrenj6: I'm thinking it's a ruby version thing - thats ruby 2.4, I can add `rescue StandardError => e` and then dippy-bot will complain too. Ill just make it explicit to make hakiri happy.

slemrmartin (approved):

lib/topological_inventory/ansible_tower/operations/worker.rb
slemrmartin: it would be best to remove this rule at all, because `rubocop -a` is causing syntax error, right?

syncrou (approved):
LGTM 👍

PR #131, opened 2020-09-08
Operations/Targeted refresh refactoring

slemrmartin:
There are lots of duplicates accross operations/targeted refresh mainly related to Sources API calls.  
Also topological api client is mixin in opposite to sources api client as a class.  
And the Operations file structure is not much consistent as it evolved in time. 

This PR unifies these issues

---

**Depends on**:

- [x] https://github.com/RedHatInsights/topological_inventory-providers-common/pull/52
- [x] https://github.com/RedHatInsights/topological_inventory-providers-common/pull/53

----

[RHCLOUD-9343](https://issues.redhat.com/browse/RHCLOUD-9343)


lib/topological_inventory/ansible_tower/operations/applied_inventories/request.rb
```diff
@@ -0,0 +1,51 @@
+require "topological_inventory-api-client"
+require "topological_inventory/providers/common/mixins/topology_api"
+require "topological_inventory/ansible_tower/operations/applied_inventories/parser"
+
+module TopologicalInventory
+  module AnsibleTower
+    module Operations
+      module AppliedInventories
+        class Request
```
slemrmartin, line 9: This code is moved from ServiceOffering.applied_inventories

lib/topological_inventory/ansible_tower/operations/ordering/order.rb
```diff
@@ -51,15 +66,16 @@ def order
           end
 
           def ansible_tower_client(source_id, task_id, identity)
-            Core::AnsibleTowerClient.new(source_id, task_id, identity)
+            Ordering::AnsibleTowerClient.new(source_id, task_id, identity)
           end
 
           # Type defined by collector here:
-          # lib/topological_inventory/ansible_tower/parser/service_offering.rb:12
+          # lib/topological_inventory/ansible_tower/parser/ordering.rb:12
```
slemrmartin: @slemrmartin TODO: rename back

lindgrenj6: Rename back? As in rename this class to service_offering?

lib/topological_inventory/ansible_tower/operations/service_offering.rb
```diff
@@ -1,50 +1,23 @@
...
-                        else
-                          [ parser.load_job_template_inventory ].compact
-                        end
-
-          update_task(task_id, :state => "completed", :status => "ok", :context => { :applied_inventories => inventories.map(&:id) })
-        rescue StandardError => err
-          logger.error("[Task #{task_id}] AppliedInventories error: #{err}\n#{err.backtrace.join("\n")}")
-          update_task(task_id, :state => "completed", :status => "error", :context => { :error => err.to_s })
+        def order
+          Ordering::Order.new(params, identity).run
```
lindgrenj6: So I like the `AppliedInventories::Request` rename - but for this one maybe we could have it be something like `Order::Request`? That way it's at least consistent.

lib/topological_inventory/ansible_tower/operations/source.rb
```diff
@@ -25,10 +25,6 @@ def connection_check
           logger.availability_check("Failed to connect to Source id:#{source_id} - #{e.message}", :error)
           [STATUS_UNAVAILABLE, e.message]
         end
-
-        def full_hostname(endpoint)
-          endpoint.host.tap { |host| host << ":#{endpoint.port}" if endpoint.port }
-        end
```
lindgrenj6, line 31: :tada:

lib/topological_inventory/ansible_tower/parser.rb
```diff
@@ -37,7 +37,7 @@ def parse_base_item(entity)
     attr_accessor :tower_url
 
     def tower_client_class
-      TopologicalInventory::AnsibleTower::Operations::Core::AnsibleTowerClient
+      TopologicalInventory::AnsibleTower::Operations::Ordering::AnsibleTowerClient
```
lindgrenj6: I agree that I don't like the `Core` name (conflicts with topo core gem), but I'm not sure I like the `Ordering::AnsibleTowerClient` name either. 

Since it's just a regular client class maybe we could just have it be `TopologicalInventory::AnsibleTower::Operations::AnsibleTowerClient`? Open to suggestions too.

slemrmartin: hm, yeah, I put it under ordering namespace because it's tied to ordering only (at this moment), but it can change in the future, your suggestion makes sense

lib/topological_inventory/ansible_tower/operations/ordering/order.rb
slemrmartin: no, just rename this comment

lib/topological_inventory/ansible_tower/operations/ansible_tower_client.rb
```diff
@@ -0,0 +1,128 @@
+require "topological_inventory/ansible_tower/logging"
+require "topological_inventory/ansible_tower/connection_manager"
+require "topological_inventory/providers/common/mixins/sources_api"
+require "topological_inventory/providers/common/mixins/x_rh_headers"
+
+module TopologicalInventory
+  module AnsibleTower
+    module Operations
+      class AnsibleTowerClient
```
slemrmartin, line 9: This is not new class, but moved from TopologicalInventory::AnsibleTower::Operations::Core::AnsibleTowerClient. I don't know why it's not compared to original file (deleted in this PR)

lindgrenj6 (approved):

lib/topological_inventory/ansible_tower/operations/order/request.rb
```diff
@@ -51,15 +66,16 @@ def order
           end
 
           def ansible_tower_client(source_id, task_id, identity)
-            Core::AnsibleTowerClient.new(source_id, task_id, identity)
+            TopologicalInventory::AnsibleTower::Operations::AnsibleTowerClient.new(source_id, task_id, identity)
```
syncrou, line 69: I do like this class naming hierarchy better than the `Core::AnsibleTowerClient` naming.

lib/topological_inventory/ansible_tower/targeted_refresh/service_instance.rb
```diff
@@ -1,35 +1,43 @@
...
           set_connection_data!
-          logger.info("ServiceInstance#refresh - Connection set successfully")
+          logger.info_ext(operation, "Connection set successfully")
 
           # Get all tasks
           tasks, refresh_state_uuid = {}, SecureRandom.uuid
           params.to_a.each do |task|
             if task['task_id'].blank? || task['source_ref'].blank?
-              logger.warn("ServiceInstance#refresh - missing data for task: #{task}")
+              logger.warn_ext(operation, "Missing data for task: #{task}")
```
syncrou, line 40: I'm not seeing much gain between:
```
logger.warn_ext(operation, "Missing data for task: #{task}")
```
vs

```
logger.warn("ServiceInstance#refresh - missing data for task: #{task}")
```

The requirement to define `self.operation` on line 28 makes logging more complicated in my opinion.

slemrmartin: @syncrou You'll find the gain during troubleshooting in Kibana. By this, you're forced to specify the operation name, which is common to all logs in operation so you will know that all operation-related logs will be filterable by one unique pattern

syncrou: I'm not going to hold up this PR. It's that these project specific logging wrappers need to be remembered by a human.

So I get why this is done, but I'm a big fan of how. I would rather we overloaded `logger.warn` and allow an accessor to pass in the object type. I even wouldn't mind a class called `KibanaLogger` which inherits from `Logger`

meta
```
logger = KibanaLogger.new(operation)
logger.warn("problem")
```

A separate method `warn_ext` get's difficult to remember to log in a specific way, and I'm worried we're going to miss it (when) those adding to this code base start to increase.

syncrou (approved):
LGTM 👍 ( Still added my thoughts about `warn_ext` but the PR is good. )

lib/topological_inventory/ansible_tower/targeted_refresh/service_instance.rb
slemrmartin: yeah, the name isn't the best. I didn't realize this solution, I like it :+1:

PR #132, opened 2020-09-10
Add response_received! method to AnsibleTower::Collector class

lindgrenj6:
Getting a NoMethodError on stage currently since the AsyncReceiver calls this method, rather than only call it for collection it is probably just easier to make this 'collector' class just do a no-op for that method.

If there is a better way to do this let me know.

### DIFF
--- a/lib/topological_inventory/ansible_tower/collector.rb
+++ b/lib/topological_inventory/ansible_tower/collector.rb
@@ -29,6 +29,10 @@ module TopologicalInventory
         end
       end
 
+      # AsyncReceiver method that needs to be on EVERY instance of collector for receptor.
+      def response_received!
+      end
+
       private
 
       attr_accessor :connection_manager, :metrics, :tower_hostname


lindgrenj6:
cc @syncrou @slemrmartin

lib/topological_inventory/ansible_tower/targeted_refresh/service_instance.rb
```diff
@@ -55,6 +55,10 @@ def async_collecting_finished(entity_type, refresh_state_uuid, total_parts)
           logger.info("ServiceInstance#refresh: finished collecting of #{entity_type} (#{refresh_state_uuid}). Total parts: #{total_parts}")
         end
 
+        def response_received!
```
slemrmartin: good catch! What about changing the async_receiver to call it only if it reponds_to?

lindgrenj6: I thought about doing it that way - but the class inherits from the AnsibleTower Collector class so we should probably have it implemented as a no-op rather than conditionally calling it in 4 different places.

lindgrenj6: Ended up moving it into the base `AnsibleTower::Collector` class, that way the base class will catch it.

syncrou (approved):
LGTM 👍 

Happy to see this on the base class.

syncrou (approved):
LGTM 👍

PR #133, opened 2020-09-16
Move to rdkafka gem

lindgrenj6:
Changing our consumers to use librdkakfa rather than the ruby-kafka gem

\# TODO:
- [x] somehow get librdkafka in our image (went with building from source, it's quick)
- [x] ~change out consumer loop for rdkafka gem consumer~ refactoring in the miq-messaging gem


Gemfile
```diff
@@ -5,14 +5,15 @@ require File.join(Bundler::Plugin.index.load_paths("bundler-inject")[0], "bundle
...
-gem "cloudwatchlogger",   "~> 0.2.1"
+gem "cloudwatchlogger", "~> 0.2.1"
 gem "concurrent-ruby"
 gem "manageiq-loggers",   "~> 0.5.0"
 gem "manageiq-messaging", "~> 0.1.2"
 gem "more_core_extensions"
 gem "optimist"
 gem "prometheus_exporter", "~> 0.4.5"
 gem "rake", ">= 12.3.3"
+gem "rdkafka", "~> 0.8"
```
syncrou: Would we still need this if manageiq-messaging pulls it in? I'm guessing not, but also guessing this PR will include the `manageiq-messaging` branch changes also ( once that is ready. )

lindgrenj6: Yep, exactly. I'll get it updated when the new branch is ready.

lindgrenj6:
@syncrou I think we're ready to try this. I pushed a branch to my repository that removes the versioning requirement on the miq-messaging gem over in the receptor_controller-client-ruby repo, we'll probably up that to 0.2.0 eventually, but for now we can just point to my branch.

lib/topological_inventory/ansible_tower/operations/worker.rb
```diff
@@ -54,18 +50,6 @@ def process_message(message)
           message.ack
           TopologicalInventory::Providers::Common::Operations::HealthCheck.touch_file
         end
-
-        # TODO: Probably move this to common eventually, if we need it elsewhere.
```
syncrou, line 58: Awesome to see everything is handled via that gem now!

syncrou (approved):
LGTM 👍

Fryguy:
@lindgrenj6 How come I don't see a PR to manageiq-messaging?

lindgrenj6:
@Fryguy I think it's still kind of in active development, it's just living in Bill's miq-messaging `rdkafka` branch for now. Once we get it tested more you'll see a PR shortly after!

Fryguy:
Thanks!  Yeah, I think making a WIP PR sooner than later would be better actually, because then we can discuss the architecture of it (for example, my first though is, from a gem perspective, that we should support both the C extension and the Ruby extension, for flexibility, but without seeing the code, I'm not sure).

lindgrenj6:
@Fryguy @bzwei ^^^^ That is a great idea. The users could specify whether they want to use the native ruby impl or not, defaulting to rdkafka maybe.

PR #134, opened 2020-09-16
even more rdkafka changes

lindgrenj6:
Receptor client changes here: https://github.com/RedHatInsights/receptor_controller-client-ruby/pull/32

basically had to disable the wait-lock at the initialization, it never came back. 

~These are other various changes I needed to get it working. The ONLY thing that isn't working currently is it seems auto-ack is broken, even with auto_ack => true it isn't acking the messages and it re-processes the messages every time.~

cc @syncrou @slemrmartin @bzwei @gmcculloug

### DIFF
--- a/lib/topological_inventory/ansible_tower/operations/worker.rb
+++ b/lib/topological_inventory/ansible_tower/operations/worker.rb
@@ -13,6 +13,7 @@ module TopologicalInventory
 
         def run
           TopologicalInventory::AnsibleTower::ConnectionManager.start_receptor_client
+          sleep 5
           logger.info("Topological Inventory AnsibleTower Operations worker started...")
 
           client.subscribe_topic(queue_opts) do |message|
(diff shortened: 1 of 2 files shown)


lib/topological_inventory/ansible_tower/operations/worker.rb
```diff
@@ -13,6 +13,7 @@ class Worker
 
         def run
           TopologicalInventory::AnsibleTower::ConnectionManager.start_receptor_client
+          sleep 5
```
syncrou, line 16: Is this due to a slow connection time?

syncrou (approved):
Looks good. I would like to do a follow up and remove the `sleep 5` [here](https://github.com/RedHatInsights/topological_inventory-ansible_tower/pull/134/files#diff-78c91e3cadb2cb4cb850c32197706dae) as a followup if we can.

PR #136, opened 2020-09-23
Partial refresh (API calls with timestamp)

slemrmartin:
Partial refresh is adding `modified__gt=<Time in ISO8601>` param to API calls.

It's using internal scheduler so first refresh is always full, then they are mixed in interval specified by ENV variables:
- `ENV['COLLECTOR_FULL_REFRESH_FREQUENCY'] || 3600 # seconds`
- `ENV['COLLECTOR_PARTIAL_REFRESH_FREQUENCY'] || 300`
- - `ENV['COLLECTOR_POLL_TIME']` is here for compatibility

The full refresh has to be repeated time to time from several reasons:

1. Sweeping (archiving) of items which were not collected 
2. Some values don't update the `modified` timestamp in Tower (tested on 3.7.0):
- changed nodes in workflow template (i.e. from "always" to "on failure") - we're collecting them, but afaik not using them
- credentials of workflow job template nodes

---

- [x] **depends on** https://github.com/RedHatInsights/topological_inventory-orchestrator/pull/110

---

[RHCLOUD-9410](https://issues.redhat.com/browse/RHCLOUD-9410)

### DIFF
--- a/lib/topological_inventory/ansible_tower/collector/scheduler.rb
+++ b/lib/topological_inventory/ansible_tower/collector/scheduler.rb
@@ -1,0 +1,100 @@
+require 'concurrent'
+
+module TopologicalInventory
+  module AnsibleTower
+    class Collector
+      class Scheduler
+        attr_accessor :full_refresh_frequency,
+                      :partial_refresh_frequency
+
+        def self.default
+          @default ||= new
+        end
+
+        def self.configure
+          if block_given?
+            yield(default)
+          else
+            default
+          end
+        end
+
+        def initialize
+          self.full_refresh_frequency    = ENV['COLLECTOR_FULL_REFRESH_FREQUENCY'] || 120
+          self.timestamps                = Concurrent::Map.new
+          self.partial_refresh_frequency = ENV['COLLECTOR_PARTIAL_REFRESH_FREQUENCY'] || 60
+        end
+
+        def add_source(source_uid)
+          timestamps[source_uid] = {:full_refresh => {}, :partial_refresh => {}}
+        end
+
+        def remove_source(source_uid)
+          timestamps.delete(source_uid)
+        end
+
+        def do_refresh?(source_uid)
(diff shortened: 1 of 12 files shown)


slemrmartin:
@lindgrenj6 Travis is failing because of receptor-client in Gemfile. Can you look at it please?

slemrmartin:
ready for review

slemrmartin:
specs should pass after https://github.com/RedHatInsights/topological_inventory-ansible_tower/pull/138 will be merged

lib/topological_inventory/ansible_tower/collector/scheduler.rb
```diff
@@ -0,0 +1,100 @@
...
+          timestamps[source_uid][type][:started_at] = Time.now.utc
+        end
+
+        def refresh_finished(type, source_uid)
+          started_at = timestamps[source_uid].try(:[], type).try(:[], :started_at)
+          if started_at.nil? && type == :partial_refresh
+            started_at = timestamps[source_uid].try(:[], :full_refresh).try(:[], :started_at)
+          end
+
+          raise "Refresh started_at for source #{source_uid} is missing!" if started_at.nil?
```
lindgrenj6, line 93: Should we be rescuing this somewhere so the thread doesn't die?

lindgrenj6 (commented):
Just a small question, other than that looks good :+1:

lib/topological_inventory/ansible_tower/collector/scheduler.rb
slemrmartin: nice catch - fixed here: https://github.com/RedHatInsights/topological_inventory-ansible_tower/pull/136/files#diff-9f1d6ce64fb0c6b1f26fd1dd864996cfef307855d94a41d41e06cab620edb153R98

lindgrenj6 (approved):

slemrmartin:
@roliveri please review

PR #137, opened 2020-09-29
Targeted refresh: archiving deleted jobs

slemrmartin:
Targeted refresh requests jobs in batches. Jobs which are not received back are considered as deleted and sent to Ingress API as archived

---

- [x] **depends on** https://github.com/RedHatInsights/topological_inventory-core/pull/210
- [x] **depends on** https://github.com/RedHatInsights/topological_inventory-persister/pull/72
---

[RHCLOUD-9508](https://issues.redhat.com/browse/RHCLOUD-9508)

### DIFF
--- a/lib/topological_inventory/ansible_tower/targeted_refresh/service_instance.rb
+++ b/lib/topological_inventory/ansible_tower/targeted_refresh/service_instance.rb
@@ -48,12 +48,12 @@ module TopologicalInventory
             tasks[task['task_id']] = task['source_ref']
 
             if tasks.length == REFS_PER_REQUEST_LIMIT
-              refresh_part(tasks, @refresh_state_uuid, SecureRandom.uuid)
+              refresh_part(tasks)
               tasks = {}
             end
           end
 
-          refresh_part(tasks, @refresh_state_uuid, SecureRandom.uuid) unless tasks.empty?
+          refresh_part(tasks) unless tasks.empty?
 
           archive_not_received_service_instances unless on_premise?
         rescue => err
@@ -95,7 +95,7 @@ module TopologicalInventory
           is_missing
         end
 
-        def refresh_part(tasks, refresh_state_uuid, refresh_state_part_uuid)
+        def refresh_part(tasks)
           tasks_id = tasks.keys.join(' | id: ')
 
           # API request, nodes and jobs under workflow job not needed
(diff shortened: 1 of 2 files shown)


lindgrenj6:
@slemrmartin @syncrou https://github.com/ManageIQ/guides/commit/702025def07e4bffb6c8875f54b5ddf9a1cca9b4

Looks like the rubocop configs got moved to manageiq-style now, ugh.

slemrmartin:
@lindgrenj6 @syncrou if we need to change all repos, I'm for moving the .rubocop-base.yml to some RedHatInsights org's repo

lindgrenj6:
I agree, that way we can tweak as we want to as well.

lib/topological_inventory/ansible_tower/targeted_refresh/service_instance.rb
```diff
@@ -34,7 +36,10 @@ def refresh
           logger.info_ext(operation, "Connection set successfully")
 
           # Get all tasks
-          tasks, refresh_state_uuid = {}, SecureRandom.uuid
+          tasks, @refresh_state_uuid = {}, SecureRandom.uuid
```
lindgrenj6, line 39: If we're going to make `refresh_state_uuid` an instance variable - maybe we don't need to pass it around between functions anymore?  Mostly talking about the `refresh_part_*` and `async_save_inventory`/`async_collecting_finished` methods.

lindgrenj6 (commented):
Just one small question - not a deal breaker by any means just an observation.

lib/topological_inventory/ansible_tower/targeted_refresh/service_instance.rb
slemrmartin: The async methods receives them from AsyncReceiver (shared with full on-premise refresh), so the param has to be there, using `@` var if there is param of the same name will be confusing. 
Not sure what do you mean by the `refresh_part_*` thing

slemrmartin: @lindgrenj6 btw do you want I refactor it? Just saying the complete retesting is quite time-consuming

lindgrenj6: Well I guess that is my question - I'm curious why we need to pass `@refresh_state_uuid` to the `refresh_part` method - since it is an instance variable now can't we just access it via `@refresh_state_uuid` instead of using the parameter?

slemrmartin: we can use it directly, so ok, I'll change it

slemrmartin:
@lindgrenj6 do you want a 2nd review or are you ok to merge it?

lindgrenj6 (approved):

PR #138, opened 2020-10-01
Use Async Operations Worker class for Source.availability_check messages

lindgrenj6:
*Depends on*: 
- [x] https://github.com/RedHatInsights/topological_inventory-providers-common/pull/55 (add class)
- [x] https://github.com/RedHatInsights/topological_inventory-providers-common/pull/56 (release v1.0.12)

----

https://issues.redhat.com/browse/RHCLOUD-9652

The client-side changes for this bug. Basically instantiate an `AsyncWorker` and then send any messages that are `Source#availability_check` over there so any ordering/applied inventories messages don't get blocked by all of the availability check messages. 

We can add any messages we want to the `ASYNC_MESSAGES` array constant, so we can really dial it down to spend the most time on what needs to be processed immediately.

cc @syncrou @slemrmartin

### DIFF
--- a/lib/topological_inventory/ansible_tower/operations/worker.rb
+++ b/lib/topological_inventory/ansible_tower/operations/worker.rb
@@ -22,7 +22,7 @@ module TopologicalInventory
             log_with(message.payload&.fetch_path('request_context', 'x-rh-insights-request-id')) do
               if ASYNC_MESSAGES.include?(message.message)
                 logger.debug("Queuing #{message.message} message for asynchronous processing...")
-                async_queue << message
+                async_worker.enqueue(message)
               else
                 model, method = message.message.to_s.split(".")
                 logger.info("Received message #{model}##{method}, #{message.payload}")
@@ -35,7 +35,7 @@ module TopologicalInventory
           logger.error("#{err.cause}\n#{err.backtrace.join("\n")}")
         ensure
           client&.close
-          async_worker&.exit
+          async_worker&.stop
           TopologicalInventory::AnsibleTower::ConnectionManager.stop_receptor_client
         end
 
@@ -50,11 +50,7 @@ module TopologicalInventory
         end
 
         def async_worker
-          @async_worker ||= TopologicalInventory::Providers::Common::Operations::AsyncWorker.new(async_queue, Processor).run
-        end
-
-        def async_queue
-          @async_queue ||= Queue.new
+          @async_worker ||= TopologicalInventory::Providers::Common::Operations::AsyncWorker.new(Processor)
         end
 
         def process_message(message)
@@ -71,7 +67,7 @@ module TopologicalInventory
 
         def start_workers
           TopologicalInventory::AnsibleTower::ConnectionManager.start_receptor_client
-          async_worker
+          async_worker.start
         end
       end
     end
--- a/lib/topological_inventory/ansible_tower/operations/worker.rb
+++ b/lib/topological_inventory/ansible_tower/operations/worker.rb
@@ -4,6 +4,7 @@ require "topological_inventory/ansible_tower/operations/source"
 require "topological_inventory/ansible_tower/connection_manager"
 require "topological_inventory/ansible_tower/messaging_client"
 require "topological_inventory/providers/common/operations/health_check"
+require "topological_inventory/providers/common/operations/async_worker"
 
 module TopologicalInventory
   module AnsibleTower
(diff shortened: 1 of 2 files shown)


lib/topological_inventory/ansible_tower/operations/worker.rb
```diff
@@ -40,6 +49,14 @@ def queue_opts
           TopologicalInventory::AnsibleTower::MessagingClient.default.worker_listener_queue_opts
         end
 
+        def async_worker
+          @async_worker ||= TopologicalInventory::Providers::Common::Operations::AsyncWorker.new(async_queue, Processor).run
```
slemrmartin: it should return instance of async worker instead of thread ID

lib/topological_inventory/ansible_tower/operations/worker.rb
```diff
@@ -4,29 +4,38 @@
...
+
+                process_message(message)
+              end
             end
           end
         rescue => err
           logger.error("#{err.cause}\n#{err.backtrace.join("\n")}")
         ensure
           client&.close
+          async_worker&.exit
```
slemrmartin: `async_worker&.stop` will be better as this variable should be an instance of AsyncWorker

lib/topological_inventory/ansible_tower/operations/worker.rb
```diff
@@ -4,29 +4,38 @@
...
           logger.info("Topological Inventory AnsibleTower Operations worker started...")
 
           client.subscribe_topic(queue_opts) do |message|
             log_with(message.payload&.fetch_path('request_context', 'x-rh-insights-request-id')) do
-              model, method = message.message.to_s.split(".")
-              logger.info("Received message #{model}##{method}, #{message.payload}")
-              process_message(message)
+              if ASYNC_MESSAGES.include?(message.message)
+                logger.debug("Queuing #{message.message} message for asynchronous processing...")
+                async_queue << message
```
slemrmartin: Can you move it to `async_worker.put` or something like that?

lib/topological_inventory/ansible_tower/operations/worker.rb
```diff
@@ -40,6 +49,14 @@ def queue_opts
           TopologicalInventory::AnsibleTower::MessagingClient.default.worker_listener_queue_opts
         end
 
+        def async_worker
+          @async_worker ||= TopologicalInventory::Providers::Common::Operations::AsyncWorker.new(async_queue, Processor).run
+        end
+
+        def async_queue
```
slemrmartin: The Queue can be part of AsyncWorker

slemrmartin (changes requested):
I was wondering about inheriting whole Worker from providers-common, but this is also ok. 
The idea is great, just a few comments

lib/topological_inventory/ansible_tower/operations/worker.rb
lindgrenj6: Oooh yeah I like that idea, changed the arity of the constructor to make the `queue` arg optional.

lib/topological_inventory/ansible_tower/operations/worker.rb
lindgrenj6: Changed it the `AsyncWorker#enqueue`

slemrmartin (approved):
:+1: LGTM

lib/topological_inventory/ansible_tower/operations/worker.rb
```diff
@@ -4,29 +4,38 @@
...
   module AnsibleTower
     module Operations
       class Worker
         include Logging
 
+        ASYNC_MESSAGES = %w[Source.availability_check].freeze
+
         def run
-          TopologicalInventory::AnsibleTower::ConnectionManager.start_receptor_client
-          sleep 5
```
roliveri, line 16: Is there no longer a need for this `sleep 5` after `start_receptor_client` in `start_workers`?

syncrou (commented):
This looks good, except for requested feedback from @roliveri.

Once that is cleared up I think this is good to go.

lib/topological_inventory/ansible_tower/operations/worker.rb
lindgrenj6: No I don't think so - there isn't a start-up delay with rdkafka like there was with the ruby-kafka library. 

With ruby-kafka it would do some message producing to let us know the client was ready to process messages etc. Since the new rdkafka library is a FFI C library the start-up time is basically nil. I can add a small timeout if we run into issues - but initially I just left it in there because we had a timeout before, it has been behaving fine locally in my testing.

roliveri (approved):

PR #142, opened 2020-11-04
Custom metrics for collector and targeted refresh

slemrmartin:
**Issue** https://github.com/RedHatInsights/topological_inventory-api/issues/320

Metrics:
* `topological_inventory_ansible_tower_collector` and `topological_inventory_ansible_tower_targeted_refresh`:
  * _error (args: type) [counter]

* **Also fixes on-premise archiving**

- [x] **depends on** #141 
---

[RHCLOUD-9945](https://issues.redhat.com/browse/RHCLOUD-9945)
[RHCLOUD-9948](https://issues.redhat.com/browse/RHCLOUD-9948)


bin/ansible_tower-collector
```diff
@@ -121,3 +120,9 @@ Signal.trap("TERM") do
   end
   exit
 end
+
+if args[:config].nil?
+  collector.collect!
```
lindgrenj6, line 123: Do we need to remove the `collector.collect!` on line 110 as well?

lib/topological_inventory/ansible_tower/targeted_refresh/service_instance.rb
```diff
@@ -57,6 +56,8 @@ def refresh
 
           archive_not_received_service_instances unless on_premise?
         rescue => err
+          metrics_err_type = on_premise? ? :receptor : :cloud unless on_premise?.nil?
```
lindgrenj6: I don't think we need the `unless on_premise?.nil?` since nil will be falsey and fall back to `:cloud` in this case. 

Unless there is a very specific case we want it to fall back to `:general` that I'm missing.

lib/topological_inventory/ansible_tower/targeted_refresh/worker.rb
```diff
@@ -48,17 +54,17 @@ def process_message(message)
...
-          raise
+          metrics&.record_error
         rescue => err
           tasks_id = if payload
                        ids = payload['params'].to_a.collect { |task| task['task_id'] }
                        ids.compact!
                      end
           logger.error("#{model}##{method} - Task[ id: #{tasks_id.to_a.join(' | id: ')} ] #{err.class.name}\n#{err.message}\n#{err}\n#{err.backtrace.join("\n")}")
-          raise
+          metrics&.record_error
```
lindgrenj6, line 67: Should we still `raise` after recording? Not sure why we'd remove it otherwise.

lib/topological_inventory/ansible_tower/targeted_refresh/service_instance.rb
slemrmartin: @lindgrenj6 yes, it's an edge case when endpoint is missing

lib/topological_inventory/ansible_tower/targeted_refresh/worker.rb
slemrmartin: I removed it from all operations. I don't know why it was there. It means, if `raise` is present, the pod ends each time there is an error in processing message.

lib/topological_inventory/ansible_tower/targeted_refresh/service_instance.rb
slemrmartin: hm, no, it still can't return nil, I'll remove it

lindgrenj6 (approved):

PR #147, opened 2020-12-01
Add `group.instance.id` settings to targeted refresh listener

lindgrenj6:
So after asking Chris Mitchell to look at our targeted collector topic on Kafka we found out consumers are getting kicked out and rebalances are occurring. That's why the listeners just stop and then start again after the rebalance is done. 

I spoke with Chris and Bill about a potential solution, and that is setting up the [group instance id](https://kafka.apache.org/documentation/#group.instance.id). Which tells Kafka that we're a consistent group of consumers, and that if we go away for a bit we'll definitely be back shortly. So do not do a rebalance (which can take minutes). It may solve our problem, as seen here:
```
[2020-12-01 20:21:37,445] INFO [GroupCoordinator 0]: Static member Some(mewtwo) with unknown member id rejoins, assigning new member id mewtwo-1606853869276, while old member mewtwo-1606853869276 will be removed. (kafka.coordinator.group.GroupCoordinator)
[2020-12-01 20:21:37,445] INFO [GroupCoordinator 0]: Static member joins during Stable stage will not trigger rebalance. (kafka.coordinator.group.GroupCoordinator)
```
This is what happens when I kill my local listener and reconnect. So the basic flow is we increase our time-out to 120 seconds so it won't rebalance until a consumer has been gone for 2 minutes, which should only happen when there is a redeployment. 

I think this is worth trying - because we're seeing consistent rebalances again like we were before where the listener isn't heart-beating right etc which is unfortunate, but this would fix that I hope.

### DIFF
--- a/lib/topological_inventory/ansible_tower/messaging_client.rb
+++ b/lib/topological_inventory/ansible_tower/messaging_client.rb
@@ -12,7 +12,13 @@ module TopologicalInventory
       end
 
       def targeted_refresh_listener
-        @targeted_refresh_listener ||= ManageIQ::Messaging::Client.open(targeted_refresh_listener_opts)
+        @targeted_refresh_listener ||= ManageIQ::Messaging::Client.open(targeted_refresh_listener_opts).tap do |client|
+          # persistent workers by pod hostname, this will prevent rebalances.
+          client.send(:kafka_client)[:"group.instance.id"] = ENV['HOSTNAME']
+
+          # 30 second timeout, after this the worker of the old hostname (ie in the event of a redeploy) will be removed and the topic will be rebalanced.
+          client.send(:kafka_client)[:"session.timeout.ms"] = 30 * 1000
+        end
       end
 
       def worker_listener_queue_opts
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,3 +1,5 @@
 Dockerfile
 Gemfile.lock
 .git
+vendor/*
+.bundle/*


lib/topological_inventory/ansible_tower/messaging_client.rb
```diff
@@ -12,7 +12,13 @@ def worker_listener
       end
 
       def targeted_refresh_listener
-        @targeted_refresh_listener ||= ManageIQ::Messaging::Client.open(targeted_refresh_listener_opts)
+        @targeted_refresh_listener ||= ManageIQ::Messaging::Client.open(targeted_refresh_listener_opts).tap do |client|
+          # persistent workers by pod hostname, this will prevent rebalances.
+          client.send(:kafka_client)[:"group.instance.id"] = ENV['HOSTNAME']
```
syncrou, line 17: So even though the hostname is different between each pod ( if we have more than one scaled up ) this stil helps us?  Or after a restart the pod has a different hostname correct? - So wouldn't a rebalance automatically occur?

lindgrenj6: Yes, a rebalance would still occur on a deploy, there is basically no way to get around that unless we set the hostname manually somehow. 

This just takes care of the workers timing out and disconnecting which is triggering a rebalance as we're seeing right now. Basically this prevents rebalancing on pod restarts due to OOM errors or due to just errors on our side.

PR #148, opened 2020-12-02
Targeted refresh: Kafka rebalance handling and messages skipping

slemrmartin:
1) Closes and creates new kafka client when Rdkafka::RdkafkaError occurs (in our case `Rdkafka::RdkafkaError\nLocal: Maximum application poll interval (max.poll.interval.ms) exceeded (max_poll_exceeded)`)

2) Skips messages without or with old "sent_at" timestamp, there is no risk of lost data. 

- [ ] **depends on** https://github.com/RedHatInsights/topological_inventory-scheduler/pull/11
---

https://issues.redhat.com/browse/RHCLOUD-10859

### DIFF
--- a/lib/topological_inventory/ansible_tower/messaging_client.rb
+++ b/lib/topological_inventory/ansible_tower/messaging_client.rb
@@ -11,14 +11,13 @@ module TopologicalInventory
         @worker_listener ||= ManageIQ::Messaging::Client.open(worker_listener_opts)
       end
 
-      def targeted_refresh_listener
-        @targeted_refresh_listener ||= ManageIQ::Messaging::Client.open(targeted_refresh_listener_opts).tap do |client|
-          # persistent workers by pod hostname, this will prevent rebalances.
-          client.send(:kafka_client)[:"group.instance.id"] = ENV['HOSTNAME']
-
-          # 30 second timeout, after this the worker of the old hostname (ie in the event of a redeploy) will be removed and the topic will be rebalanced.
-          client.send(:kafka_client)[:"session.timeout.ms"] = 30 * 1000
+      def targeted_refresh_listener(renew: false)
+        if renew
+          @targeted_refresh_listener&.close
+          @targeted_refresh_listener = nil
         end
+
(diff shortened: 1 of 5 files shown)


lib/topological_inventory/ansible_tower/messaging_client.rb
```diff
@@ -11,14 +11,13 @@ def worker_listener
         @worker_listener ||= ManageIQ::Messaging::Client.open(worker_listener_opts)
       end
 
-      def targeted_refresh_listener
-        @targeted_refresh_listener ||= ManageIQ::Messaging::Client.open(targeted_refresh_listener_opts).tap do |client|
-          # persistent workers by pod hostname, this will prevent rebalances.
-          client.send(:kafka_client)[:"group.instance.id"] = ENV['HOSTNAME']
```
slemrmartin, line 17: Could this cause current frozen prod pods? 
Anyway it didn't solve the problem so I'm reverting it

lib/topological_inventory/ansible_tower/messaging_client.rb
```diff
@@ -11,14 +11,13 @@ def worker_listener
...
-
-          # 30 second timeout, after this the worker of the old hostname (ie in the event of a redeploy) will be removed and the topic will be rebalanced.
-          client.send(:kafka_client)[:"session.timeout.ms"] = 30 * 1000
```
slemrmartin, line 20: the problem is related to `max.poll.interval.ms`, so I believe this can't help too

lib/topological_inventory/ansible_tower/messaging_client.rb
lindgrenj6: Yeah I'm not sure, this should just prevent rebalances, but obviously it didn't work so I'm ok reverting it.

lib/topological_inventory/ansible_tower/messaging_client.rb
lindgrenj6: Yeah I'm not sure whats going on why they would just freeze like this. This would increase the time a client has to process a message etc, I'm ok with reverting it for now. But in the future if we're seeing constant disconnects/reconnects this setting might help us.

lindgrenj6 (approved):

